PR #30, opened 2023-01-11
Feature: Mark the plugin as wordpress plugin in Composer for easier installation

michielgerritsen:
Hi,

When trying to install this plugin using Composer, it is not marked as a WordPress plugin, which causes it to get placed in the `vendor` directory instead of the `wp-content/plugins` directory. This PR fixes this.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -1,6 +1,7 @@
 {
   "name": "postnl/postnl-for-woocommerce",
   "license": "GPL-3.0",
+  "type": "wordpress-plugin",
   "require": {
 	"clegginabox/pdf-merger": "dev-master"
   },


iyut:
Hi @michielgerritsen , may I know the steps on how we test this?

michielgerritsen:
@iyut Yeah sure. Take this `composer.json` for example:

```json
{
    "name": "controlaltdelete/wordpress-type-example",
    "require": {
        "postnl/postnl-for-woocommerce": "@dev"
    },
    "repositories": [
        {
            "type": "github",
            "url": "git@github.com:Progressus-io/postnl-for-woocommerce.git"
        }
    ],
    "extra": {
        "installer-paths": {
            "wp-content/plugins/{$name}/": [
                "type:wordpress-plugin"
            ]
        }
    },
    "minimum-stability": "dev"
}
```

If you now run this command, all files will end up in the `vendor` directory:
```
composer require postnl/postnl-for-woocommerce:@dev
```

If this PR is merged, these files will be placed in `wp-content/plugins/postnl-for-woocommerce`.

This PR allows us to manage the plugins through Composer, so we don't need to keep a version in git. Other plugins support this also:
- [Yoast](https://github.com/Yoast/wordpress-seo/blob/trunk/composer.json#L17)
- [WP Rocket](https://github.com/wp-media/wp-rocket/blob/develop/composer.json#L19)

iyut:
Just tested it. yeah the changes looks good to me!

iyut (approved):
The changes LGTM!

PR #33, opened 2023-01-19
Load autoloader file with full path

Abdalsalaam:


iyut (approved):
the changes looks good to me!

PR #36, opened 2023-01-25
Fix/merge labels in bulk - Use the same barcode if it's created within 7 days.

iyut:
### Description

Currently, the plugin will always create a new barcode whenever the label is created. This PR will prevent that. It will check if there is any existing barcode within 7 days and use that instead.

### Steps to Test

1. Create a couple of new order.
2. Create a new shipping label on one of the order.
3. Go to orders list in the admin and check a couple of orders that you just created.
4. Choose "PostNL Create Label" in the bulk action and click "Submit" button.
5. When the label creation is finished, you will notice that the shipping label will still use the same barcode.

### DIFF
--- a/src/Order/Base.php
+++ b/src/Order/Base.php
@@ -374,6 +374,11 @@ abstract class Base {
 
 		$saved_data['labels'] = array_merge( $labels, $return_labels );
 		*/
+		$saved_data['barcode'] = array(
+			'value'      => $barcode,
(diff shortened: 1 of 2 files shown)


Abdalsalaam (approved):
Looks good and works as expected

iyut:
Thanks for reviewing the PR @Abdalsalaam !

PR #37, opened 2023-01-25
Fix cart ( non-standard delivery fee ) name and support multiple types

Abdalsalaam:
### Description

Currently, the plugin will always show "PostNL Evening Fee" as a cart fee even if you select Morning delivery. This PR fix it and will show the correct fee name.

### Steps to Test

1. In checkout, use a valid Netherlands shipping address.
2. Select Delivery Day and choose Morning delivery
3. Check the fee name in the cart totals table.


iyut (approved):
The changes works well and looks good!

PR #38, opened 2023-01-31
Feature/Free shipping option

Abdalsalaam:
### Description

Free shipping can be provided based on the cart total amount

### Steps to Test

1. Set (  Minimum amount for free shipping  ) for the shipping method from Shipping > manage shipping methods > PostNL
2. Add any product to the cart and make sure the total less than the minimum
3. Check shipping rates provided, it should be paid shipping
4. Then update the cart and let the total exceed the minimum and check the shipping rate, it should be free for PostNL.

### DIFF
--- a/src/Shipping_Method/PostNL.php
+++ b/src/Shipping_Method/PostNL.php
@@ -90,7 +90,7 @@ class PostNL extends \WC_Shipping_Flat_Rate {
 		$form_fields['minimum_for_free_shipping'] = array(
 			'title' 		=> esc_html__( 'Minimum amount for free shipping', 'postnl-for-woocommerce' ),
 			'type' 			=> 'number',
-			'label' 		=> esc_html__( '0 or keep it empty to disable this feature', 'postnl-for-woocommerce' ),
+			'description'   => esc_html__( '0 or keep it empty to disable this feature', 'postnl-for-woocommerce' ),
 			'default' 		=> 0,
 		);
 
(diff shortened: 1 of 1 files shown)


src/Shipping_Method/PostNL.php
```diff
@@ -49,19 +49,51 @@ public function postnl_init() {
...
+			);
+
+			$this->add_rate( $rate );
+		} else {
+			parent::calculate_shipping();
+		}
+	}
+
 	/**
 	 * Manipulate the title default.
```
iyut, line 79:
```suggestion
	 * Add form fields for PostNL.
```
Let's update the comment as well.

src/Shipping_Method/PostNL.php
```diff
@@ -49,19 +49,51 @@ public function postnl_init() {
...
 	 *
 	 * @param Array $form_fields List of instance form fields.
 	 *
 	 * @return Array
 	 */
-	public function change_title_default( $form_fields ) {
+	public function instance_form_fields( $form_fields ) {
+		// Change title default value
```
iyut, line 86:
```suggestion
		// Change title default value.
```

src/Shipping_Method/PostNL.php
```diff
@@ -49,19 +49,51 @@ public function postnl_init() {
...
 		$form_fields['title']['default'] = $this->method_title;
+
+		// Minimum for free shipping
```
iyut, line 89:
```suggestion
		// Minimum for free shipping.
```

iyut (approved):
Code wise, it looks good to me. But I left some comment regarding the comment.

PR #48, opened 2023-02-16
Remove globalpack barcode types validation

Abdalsalaam:
### Description

 Remove globalpack barcode types validation 

### Steps to Test

1. Change the  (GlobalPack Barcode Type) in the plugin setting : Shipping > manage shipping methods > PostNL, and use any random letters ( To test the wrong barcode scenario ).
2. Create an international order and choose PostNL as a shipping method.
3. As an Admin, try to generate the shipping label for the order, An error message should appear.
4. Change the (GlobalPack Barcode Type) to 'CD' or any other barcode type and retry the previews step.


iyut:
Is the client confirm that there is no validation needed for this globalpack barcode type?

Abdalsalaam:
@iyut Yes.

iyut (approved):
LGTM!

PR #52, opened 2023-03-09
remove unused function

Abdalsalaam:
this code not used anywhere, and this will block the JS code if select2 not enqueued


iyut (approved):
looks good to me!

PR #53, opened 2023-03-22
Fix/label creation fatal error

Abdalsalaam:
### Description
Fix fatal error when trying to create label for order with **deleted product**.

### Steps to Test
1. Create an Order with PostNL as the shipping method.
2. Delete one of the order _WooCommerce products_ and try to create the label.

### DIFF
--- a/src/Rest_API/Shipping/Item_Info.php
+++ b/src/Rest_API/Shipping/Item_Info.php
@@ -223,7 +223,7 @@ class Item_Info extends Base_Info {
 		foreach ( $order->get_items() as $item_id => $item ) {
 			$product = $item->get_product();
 
-			if ( empty( $product ) || $product->is_virtual() ) {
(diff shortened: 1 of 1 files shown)


src/Rest_API/Shipping/Item_Info.php
```diff
@@ -223,7 +223,7 @@ public function convert_data_to_args( $post_data ) {
 		foreach ( $order->get_items() as $item_id => $item ) {
 			$product = $item->get_product();
 
-			if ( $product->is_virtual() ) {
+			if ( empty( $product ) || $product->is_virtual() ) {
```
iyut, line 226:
```suggestion
			if ( ! is_a( $product, 'WC_Product' ) || $product->is_virtual() ) {
```

iyut (changes requested):
it might be good to make sure if the `$product` is indeed a `WC_Product` object.

PR #56, opened 2023-03-25
Feature/associated shipping methods

Abdalsalaam:
### Description
Add an field to select which Shipping Methods can be associated with PostNL and default to PostNL.

### Steps to Test
1. Edit plugin settings from WooCommerce > Settings > Shipping > PostNL, Under "Checkout Settings" you will find "Shipping Methods" fill it with associated methods and save the settings.
2. After adding products to the cart, and selecting any associated method as a shipping method for the cart, PostNL fields should be displayed.
3. Also PostNL metabox should be displayed in the admin dashboard order edit screen.

### DIFF
--- a/src/Shipping_Method/Settings.php
+++ b/src/Shipping_Method/Settings.php
@@ -229,6 +229,14 @@ class Settings extends \WC_Settings_API {
 				'type'        => 'title',
 				'description' => esc_html__( 'Please configure your checkout preferences.', 'postnl-for-woocommerce' ),
 			),
+			'supported_shipping_methods' => array(
+				'title'       => esc_html__( 'Shipping Methods', 'postnl-for-woocommerce' ),
+				'type'        => 'multiselect',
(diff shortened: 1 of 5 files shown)


src/Shipping_Method/Settings.php
```diff
@@ -1191,4 +1199,26 @@ public function get_enable_logging() {
...
+	}
+
+	/**
+	 * Get supported shipping methods from the settings.
+	 *
+	 * @return array.
+	 */
+	public function get_supported_shipping_methods() {
+		$suppoted_shipping_methods = (array) $this->get_option( 'supported_shipping_methods' );
+		// Add PostNL method by default
```
iyut, line 1219:
```suggestion
		// Add PostNL method by default.
```
Every comment should end with dot (.)

iyut (approved):
LGTM! have added some comment for fixing some spacing. Can you fix those first before merging it?

PR #61, opened 2023-03-29
Feature/create label action button

Abdalsalaam:
### Description
Adding "Create & Download  Label" actions buttons on the orders list page.

### Steps to Test
1. On the orders list page, you will find PostNL actions buttons if the order contains PostNL as a shipment method.
2. If the order doesn't have a label the action button will be "Create Label".
3. After clicking on the "Create Label" action button & the label created successfully, the action button will be "Print Label".


src/Order/Bulk.php
```diff
@@ -340,4 +344,72 @@ public function render_messages() {
...
+				$array_messages[] = array(
+					'message' => sprintf( esc_html__( '#%1$s : PostNL label has been created.', 'postnl-for-woocommerce' ), $order_id ),
+					'type'    => 'success',
+				);
+			} catch ( \Exception $e ) {
+				$array_messages[] = array(
+					'message' => sprintf( '#%1$s : %2$s', $order_id, $e->getMessage() ),
+					'type'    => 'error',
+				);
+			}
```
iyut, line 407: It seems this code is redundant with [this code](https://github.com/Progressus-io/postnl-for-woocommerce/blob/d77024ab20addc5940dce4baf1a723fd50966009/src/Order/Bulk.php#L81-L108) under `process_order_bulk_actions()` method. Do you think we can make new method for this?

iyut (changes requested):
I have added a comment about redundant code. Can you check it?

src/Order/Bulk.php
Abdalsalaam: @iyut Yes, that what I thought, I'm thinking of making a method that return array of messages & labels, what do you  think ? do you have better idea?

iyut: > I'm thinking of making a method that return array of messages & labels
I think this is a good idea! Let's do that!

PR #65, opened 2023-04-05
Possibility to add Shipping Option by Default to all orders

sal4sup:


src/Order/Base.php
```diff
@@ -107,8 +111,8 @@ public function meta_box_fields() {
 					'label'         => __( 'ID Check: ', 'postnl-for-woocommerce' ),
 					'placeholder'   => '',
 					'description'   => '',
-					'value'         => '',
-					'show_in_bulk'  => true,
+                    'value'         => in_array('id_check', $default_shipping_options) ? 'yes' : '',
```
Abdalsalaam, line 114: Use tabs instead of spaces please.

src/Order/Base.php
```diff
@@ -119,8 +123,8 @@ public function meta_box_fields() {
 					'label'         => __( 'Insured Shipping: ', 'postnl-for-woocommerce' ),
 					'placeholder'   => '',
 					'description'   => '',
-					'value'         => '',
-					'show_in_bulk'  => true,
+                    'value'         => in_array('insured_shipping', $default_shipping_options) ? 'yes' : '',
```
Abdalsalaam, line 126: same thing here

src/Order/Base.php
```diff
@@ -92,6 +92,10 @@ function( $field ) {
 	 * List of meta box fields.
 	 */
 	public function meta_box_fields() {
+        // Get the default shipping options.
+        $settings = get_option( 'woocommerce_postnl_settings' );
+        $default_shipping_options = isset( $settings['default_shipping_options'] ) ? $settings['default_shipping_options'] : array();
+        //   $default_shipping_options = $this->settings->get_default_shipping_options();
```
Abdalsalaam, line 98: Why you didn't use `get_default_shipping_options` function ?

src/Shipping_Method/Settings.php
```diff
@@ -1191,4 +1213,17 @@ public function get_enable_logging() {
...
 	}
+
+    /**
+     * Get all shipping options.
+     *
+     * @return array
+     */
+    public function get_default_shipping_options() {
+        $settings = get_option( 'woocommerce_postnl_settings' );
+        $default_shipping_options = isset( $settings['default_shipping_options'] ) ? $settings['default_shipping_options'] : array();
```
Abdalsalaam, line 1224: Use the pre built function to get the option

```suggestion
return $this->get_country_option( 'default_shipping_options' );
```

Abdalsalaam (changes requested):
Not all of these options can be combined together, how you will limit these combinations ?

PR #86, opened 2023-04-28
Fix order note

sal4sup:


src/Order/Single.php
```diff
@@ -456,12 +456,12 @@ public function save_meta_box_ajax() {
...
 			$this->delete_label_files( $labels );
 
-			if ( $this->settings->is_woocommerce_email_enabled() && ! empty( $tracking_note ) ) {
+			if ( ! empty( $tracking_note ) ) {
 				$return_data = array_merge(
 					$result['saved_data'],
 					array(
 						'tracking_note' => $tracking_note,
-						'note_type'     => Utils::get_tracking_note_type(),
+						'note_type'     => $this->settings->is_woocommerce_email_enabled() ? Utils::get_tracking_note_type() : Utils::get_private_tracking_note_type(),
```
Abdalsalaam, line 464: I suggest to remove these functions scene we don't use it anywhere else 
`'note_type'     => $this->settings->is_woocommerce_email_enabled() ? 'customer' : 'private',`

what do you think?

Abdalsalaam (changes requested):

src/Order/Single.php
sal4sup: I agree

PR #107, opened 2023-06-27
Fix checkout php warnings

Abdalsalaam:
A customer has filed a complaint about the following error showing up in their checkout:
image

This is a compatibility issue with a checkout modifier


iyut (approved):
LGTM!

PR #123, opened 2023-07-07
Fix fetal error if no PDFs to merge on bulk

Abdalsalaam:
task #861myfphx

### DIFF
--- a/src/Order/Bulk.php
+++ b/src/Order/Bulk.php
@@ -81,14 +81,14 @@ class Bulk extends Base {
 
 		if ( ! empty( $object_ids ) ) {
 			foreach ( $object_ids as $order_id ) {
-				$result           = $this->generate_label_and_notes( $order_id, $_REQUEST );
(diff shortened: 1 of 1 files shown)


src/Order/Bulk.php
```diff
@@ -83,7 +83,10 @@ public function process_order_bulk_actions( $redirect, $doaction, $object_ids )
 			foreach ( $object_ids as $order_id ) {
 				$result           = $this->generate_label_and_notes( $order_id, $_REQUEST );
 				$array_messages[] = $result['message'];
```
iyut, line 85: @Abdalsalaam , what do you think if we also check on `$result['message']` if it has value or not?

iyut (changes requested):

src/Order/Bulk.php
Abdalsalaam: Done, thank you !

PR #133, opened 2023-07-25
Check Insurance Amount Limit

sal4sup:


src/Rest_API/Shipping/Item_Info.php
```diff
@@ -940,4 +943,22 @@ protected function check_mailbox_weight_limit( $backend_data, $order_weight ) {
...
+	 * @param $backend_data  .
+	 * @param $order_total  .
+	 *
+	 * @return void.
+	 * @throws \Exception if the order weight exceeds € 5000.
+	 */
+	protected function check_insurance_amount_limit( $backend_data, $order_total ) {
+		if ( 'yes' === $backend_data['insured_shipping'] && $order_total > 5000 ) {
+			throw new \Exception(
+				__( 'Insurance amount  is required and cannot exceed the maximum allowed amount (€ 5000) Your total is. ' . $order_total, 'postnl-for-woocommerce' )
```
Abdalsalaam, line 959:
```suggestion
				__( 'Insurance amount is required and cannot exceed the maximum allowed amount (€ 5000). Your total is: ' . $order_total, 'postnl-for-woocommerce' )
```

this will works with sites using "**€**" currency, check this with the client please

Abdalsalaam (changes requested):

PR #134, opened 2023-07-26
Fix/multicollo barcode - Fixing Multicollo Barcodes

iyut:
This PR will create unique barcode for each labels if the user input number of labels more than 1.
This PR also change some variable in the create label ( `/v1/shipment` ) request.

1. Create a new order.
2. Create a PostNL label from Order edit page in admin.
3. Print the label. Each label will have unique barcode.
4. Check the log from WooCommerce >> Status >> Log. The request will show `Barcode` for each addresses array. And `MainBarcode` will always use the first barcode.
Example : 
Screenshot 2023-07-26 at 14 41 21

### Changelog entry

> * Fix multi-collo barcodes call.

### DIFF
--- a/src/Order/Base.php
+++ b/src/Order/Base.php
@@ -547,6 +547,7 @@ abstract class Base {
 		$this->delete_label( $saved_data );
 		unset( $saved_data['backend'] );
 		unset( $saved_data['labels'] );
+		unset( $saved_data['barcodes'] );
 
 		$order->update_meta_data( $this->meta_name, $saved_data );
 		$order->save();
--- a/src/Order/Base.php
+++ b/src/Order/Base.php
@@ -427,9 +427,9 @@ abstract class Base {
 			'saved_data' => $saved_data,
 		);
 
-		$barcodes                    = $this->maybe_create_multi_barcodes( $label_post_data );
-		$label_post_data['barcode']  = $barcodes[0]; // for MainBarcode.
-		$label_post_data['barcodes'] = $barcodes;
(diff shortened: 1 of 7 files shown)


src/Order/Base.php
```diff
@@ -540,7 +547,6 @@ public function delete_meta_value( $order_id ) {
 		$this->delete_label( $saved_data );
 		unset( $saved_data['backend'] );
 		unset( $saved_data['labels'] );
-		unset( $saved_data['barcode'] );
```
Abdalsalaam, line 543: @iyut  we need to delete the barcodes after deleting the label  regarding to [this requirement](https://app.clickup.com/t/861mpdmz6)

Abdalsalaam (changes requested):

src/Order/Base.php
iyut: Thanks for reviewing this PR @Abdalsalaam , so are you confirming that we dont need to saved the barcode for 7 days? Because [this code](https://github.com/Progressus-io/postnl-for-woocommerce/blob/a6a753e0271616d80cafaf62584936103743ee51/src/Order/Base.php#L620-L626) seems we need to use the saved barcode at least within 7 days.

Abdalsalaam: @iyut yes we need to save the barcode for 7 days, so if the label creation failed it will use the same barcode on the next try, but deleting the label should delete the barcode as well.

Abdalsalaam (approved):
LGTM

PR #142, opened 2023-09-25
Auto complete order  once an order has been pre-alerted and printed

sal4sup:


src/Order/Base.php
```diff
@@ -461,6 +461,12 @@ function( $label ) {
 			},
 			$labels
 		);
+
+		if($this->settings->is_auto_complete_order_enabled()){
```
Abdalsalaam, line 465:
```suggestion
		if( $this->settings->is_auto_complete_order_enabled() ) {
```

src/Shipping_Method/Settings.php
```diff
@@ -1299,4 +1308,22 @@ public function get_supported_shipping_methods() {
...
+
+	/**
+	 * Get Automatically change order status to Completed value from the settings.
+	 *
+	 * @return String
+	 */
+	public function get_auto_complete_order() {
+		return $this->get_country_option( 'auto_complete_order', '' );
+	}
+	/**
```
Abdalsalaam, line 1320:
```suggestion
	}
	
	/**
```

Abdalsalaam (changes requested):
Loogs good, I just added small changes

PR #161, opened 2024-05-22
Fix/translations

Abdalsalaam:
Fix plugin settings translation


iyut (approved):
The changes does fix the issue. And the code LGTM!

PR #165, opened 2024-06-26
check if product is exist

OlfatProgressus:
https://app.clickup.com/t/8688wz6gw
Fatal error: Uncaught Error: Call to a member function get_meta() on bool in /home/customer/www/[eluun.nl/public_html/wp-content/plugins/woo-postnl/src/Utils.php:738](http://eluun.nl/public_html/wp-content/plugins/woo-postnl/src/Utils.php:738)

this error indicates that the $product is false, meaning that wc_get_product is unable to find a product for a given item ID. This can happen if the product ID doesn't exist

### DIFF
--- a/README.md
+++ b/README.md
@@ -38,6 +38,7 @@ PostNL’s official extension for WooCommerce on WordPress. Manage your national
 
 ### 5.4.2
 * Fix: Error "Invalid nonce" when trying to delete labels.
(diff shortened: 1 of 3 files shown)


Abdalsalaam (changes requested):
nice work, we can use `is_a` to handle any other scenario

Abdalsalaam (approved):
LGTM

PR #171, opened 2024-07-17
Product editor compatibility

ferroche90:
**Enhancement: Add Custom Shipping Blocks to the WooCommerce Product Editor**
This PR adds a new Product_Editor class to the PostNL WooCommerce plugin, enhancing the product editor with custom shipping-related blocks. The blocks include:

**Letterbox Parcel Checkbox:** Enable/disable letterbox parcel.
**Max Quantity per Letterbox:** Specify max quantity per letterbox parcel.
**Country of Origin Select:** Choose the country of origin.
**HS Tariff Code Text:** Input HS Tariff Code for international shipping.
The **Product_Editor** class is instantiated in the **Main** class for seamless integration.

image

image

### DIFF
--- a/src/Main.php
+++ b/src/Main.php
@@ -11,6 +11,8 @@ if ( ! defined( 'ABSPATH' ) ) {
 	exit;
 }
 
+use PostNLWooCommerce\Product\Product_Editor;
+
 /**
  * Class Main
  *
@@ -44,6 +46,13 @@ class Main {
 	 * @var PostNLWooCommerce\Product\PostNL
 	 */
 	public $shipping_product = null;
+
+	/**
+	 * Product Editor.
+	 *
+	 * @var PostNLWooCommerce\Product\Product_Editor
+	 */
+	public $product_editor = null;
 
 	/**
 	 * Shipping Order.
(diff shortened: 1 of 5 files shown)


src/Main.php
```diff
@@ -97,6 +108,15 @@ public function declare_wc_hpos_compatibility() {
 		}
 	}
 
+	/**
+	 * Declare Product Editor compatibility.
+	 */
+	public function declare_product_editor_compatibility() {
+		if ( class_exists( '\Automattic\WooCommerce\Utilities\FeaturesUtil' ) ) {
+			\Automattic\WooCommerce\Utilities\FeaturesUtil::declare_compatibility( 'product_block_editor', 'postnl-for-woocommerce/postnl-for-woocommerce.php', true );
```
Abdalsalaam, line 116:
```suggestion
			\Automattic\WooCommerce\Utilities\FeaturesUtil::declare_compatibility( 'product_block_editor', POSTNL_WC_PLUGIN_BASENAME, true );
```

Can we use this instead, since its a published in WP.org with different folder name

Abdalsalaam (changes requested):
All looks good, I just left a comment to fix the declaration
thanks!

src/Main.php
ferroche90: @Abdalsalaam, no problem! Thank you for the suggestion!

Abdalsalaam (approved):
LGTM, can you just fix the changelog, since we didn't release 5.4.3

PR #173, opened 2024-07-22
Ignore digital products from letterbox

OlfatProgressus:


src/Utils.php
```diff
@@ -744,7 +744,8 @@ public static function check_products_for_letterbox( $products ) {
 
 			// If one of the item is not letterbox product, then the order is not eligible automatic letterbox.
 			// Thus should return false immediately.
-			if ( 'yes' !== $is_letterbox_product ) {
+			$is_digital = $product->is_virtual() || $product->is_downloadable();
```
Abdalsalaam, line 747: I suggest to check if its virtual product before getting `Product\Single::LETTERBOX_PARCEL` meta, if its virtual there is no need to get the meta value.

OlfatProgressus: yes you are right

src/Utils.php
```diff
@@ -739,6 +739,11 @@ public static function check_products_for_letterbox( $products ) {
 				// If the product is not found, consider the order not eligible.
 				return false;
 			}
+			
+			$is_digital = $product->is_virtual() || $product->is_downloadable();
+			if( !$is_digital ) {
+				return false;
```
Abdalsalaam, line 745: This will consider the order not eligible for letterbox, what we should do is ignoring this product 
```suggestion
				continue;
```

you should also keep in your mind that the order could have only one product so if if its virtual this will always consider it as a letterbox because `$total_ratio_letterbox_item` will be 0, so we might need to check if the order have eligible products or not

Abdalsalaam (changes requested):

OlfatProgressus:
@Abdalsalaam Did you checked my last update?

src/Utils.php
```diff
@@ -732,14 +732,19 @@ public static function is_eligible_auto_letterbox( $order ) {
 	 */
 	public static function check_products_for_letterbox( $products ) {
 		$total_ratio_letterbox_item = 0;
-
+		$has_letterbox_product = false;
```
Abdalsalaam, line 735:
```suggestion
		$has_letterbox_product      = false;

```

src/Utils.php
```diff
@@ -732,14 +732,19 @@ public static function is_eligible_auto_letterbox( $order ) {
...
 		foreach ( $products as $item_id => $item ) {
 			$product              = wc_get_product( $item['product_id'] ?? $item->get_product_id() );
```
Abdalsalaam, line 737:
```suggestion
			$product = wc_get_product( $item['product_id'] ?? $item->get_product_id() );
```

src/Utils.php
```diff
@@ -748,14 +753,15 @@ public static function check_products_for_letterbox( $products ) {
...
+			$has_letterbox_product 		= true;
 			$quantity                   = $item['quantity'] ?? $item->get_quantity();
 			$qty_per_letterbox          = intval( $product->get_meta( Product\Single::MAX_QTY_PER_LETTERBOX ) );
 			$ratio_letterbox_item       = 0 != $qty_per_letterbox ? 1 / $qty_per_letterbox : 0;
 			$total_ratio_letterbox_item += ( $ratio_letterbox_item * $quantity );
 		}
 
 		// If the total ratio is more than 1, that means order items cannot be packed using letterbox.
-		return ( $total_ratio_letterbox_item <= 1 ) ? true : false;
+		return ( $total_ratio_letterbox_item <= 1 && $has_letterbox_product) ? true : false;
```
Abdalsalaam, line 764:
```suggestion
		return $has_letterbox_product && $total_ratio_letterbox_item <= 1;
```

src/Utils.php
```diff
@@ -732,14 +732,19 @@ public static function is_eligible_auto_letterbox( $order ) {
...
 			$product              = wc_get_product( $item['product_id'] ?? $item->get_product_id() );
 			if ( ! is_a( $product, 'WC_Product' ) ) {
 				// If the product is not found, consider the order not eligible.
 				return false;
 			}
 
+			$is_digital = $product->is_virtual() || $product->is_downloadable();
+			if( $is_digital ) {
+				continue;
+			}
```
Abdalsalaam, line 746:
```suggestion
			if( ! $product->needs_shipping() ) {
				continue;
			}
```

Abdalsalaam (changes requested):
@OlfatProgressus  Looks good, I just added some comments to make the code better
thanks!

PR #174, opened 2024-07-25
Update logo

OlfatProgressus:


Abdalsalaam (changes requested):
I think `update-logo` branch is not created from main branch so it have so many changes not related to "Change logo", check it please.

PR #175, opened 2024-07-25
logo update

OlfatProgressus:


Abdalsalaam (changes requested):
Can we get the png logo with transparent background? Just to let it looks good with any "Container" background color.

OlfatProgressus:
I added a comment on the task for the client to send png version of the new login but he didn't reply yet

PR #178, opened 2024-08-07
Add pickup points to email

OlfatProgressus:


src/Frontend/Base.php
```diff
@@ -111,6 +111,8 @@ public function init_hooks() {
 		add_action( 'woocommerce_checkout_update_order_meta', array( $this, 'calculate_non_standard_fee' ), 20, 2 );
 		add_filter( 'postnl_frontend_checkout_tab', array( $this, 'add_checkout_tab' ), 10, 2 );
 		add_action( 'postnl_checkout_content', array( $this, 'display_content' ), 10, 2 );
+		add_filter('woocommerce_email_order_meta_fields', array( $this, 'add_pickup_points_fields_to_email'), 10, 3);
+
 	}
```
Abdalsalaam, line 116:
```suggestion
		add_filter( 'woocommerce_email_order_meta_fields', array( $this, 'add_pickup_points_fields_to_email' ), 10, 3 );
	}
```

src/Frontend/Base.php
```diff
@@ -131,6 +133,80 @@ abstract public function add_checkout_tab( $tabs, $response );
...
+		$data = $this->get_data( $order->get_id() );
+		if( ! empty( $data['frontend'] ) ){
+			$value = $this->generate_pickup_points_email_html($data['frontend']);
+			$fields['Pickup address'] = array(
+				'label' => __( 'Pick up at PostNL-point', 'postnl-for-woocommerce' ),
+				'value' => $value
+			);
+		}
+		return $fields;
+	}
```
Abdalsalaam, line 146:
```suggestion
	/**
	 * Add pickup points info to email templates.
	 *
	 * @param array     $fields Current fields.
	 * @param bool      $sent_to_admin If should sent to admin.
	 * @param \WC_Order $order Order instance.
	 *
	 * @return array
	 */
	public function add_pickup_points_fields_to_email( $fields, $sent_to_admin, $order ) {
		$data = $this->get_data( $order->get_id() );
		if ( ! empty( $data['frontend'] ) ) {
			$value                    = $this->generate_pickup_points_email_html( $data['frontend'] );
			$fields['Pickup address'] = array(
				'label' => __( 'Pick up at PostNL-point', 'postnl-for-woocommerce' ),
				'value' => $value,
			);
		}

		return $fields;
	}
```

src/Frontend/Base.php
```diff
@@ -131,6 +133,80 @@ abstract public function add_checkout_tab( $tabs, $response );
...
+			$filtered_infos['address'] = implode( ', ', $address_info );
+			ksort( $filtered_infos );
+		}
+
+		if ( empty( $filtered_infos ) ) {
+			return;
+		}
+		$value = "
+		
+		<div class='postnl-info-container pickup-points-info'>";
```
Abdalsalaam, line 185:
```suggestion
		$value = "<div class='postnl-info-container pickup-points-info'>";
```

src/Frontend/Base.php
```diff
@@ -131,6 +133,80 @@ abstract public function add_checkout_tab( $tabs, $response );
...
+		);
+
+		if ( ! empty( $address_info ) ) {
+			$filtered_infos['address'] = implode( ', ', $address_info );
+			ksort( $filtered_infos );
+		}
+
+		if ( empty( $filtered_infos ) ) {
+			return;
+		}
```
Abdalsalaam, line 182: I can see this is a duplicated code (class Single.php), we can move it to Utils instead of duplicating it since it use the same thing in both classes

Abdalsalaam (changes requested):

src/Frontend/Base.php
```diff
@@ -131,6 +132,63 @@ abstract public function add_checkout_tab( $tabs, $response );
...
+	/**
+	 * Generate the dropoff points html information.
+	 *
+	 * @param Array $infos Dropoff points informations.
+	 */
+	public function generate_pickup_points_email_html( $infos ) {
+		$filtered_infos = Utils::get_pickup_points_infos( $infos );
+		if ( empty( $filtered_infos ) ) {
+			return;
+		}
```
Abdalsalaam, line 166: this function expected to return string, but here its void return

src/Utils.php
```diff
@@ -771,4 +771,43 @@ public static function prepare_shipping_options( $selected_value ) {
...
+		);
+
+		if ( ! empty( $address_info ) ) {
+			$filtered_infos['address'] = implode( ', ', $address_info );
+			ksort( $filtered_infos );
+		}
+
+		if ( empty( $filtered_infos ) ) {
+			return;
+		}
```
Abdalsalaam, line 811: Same as the previous comment

Abdalsalaam (changes requested):

PR #182, opened 2024-09-06
Fix warnings

OlfatProgressus:
PHP Warning: Undefined array key "delivery_day_type" in/wp-content/plugins/woo-postnl/src/Frontend/Base.php on line 355\n
PHP Warning: Undefined array key "_postnl_letterbox_parcel" in /wp-content/plugins/woo-postnl/src/Product/Single.php on line 193\n
PHP Warning: Undefined array key "wc_actions" in /wp-content/plugins/woo-postnl/src/Order/OrdersList.php on line 110\n

### DIFF
--- a/src/Order/OrdersList.php
+++ b/src/Order/OrdersList.php
@@ -107,7 +107,7 @@ class OrdersList extends Base {
 	 * @return array.
 	 */
 	public function add_order_delivery_date_column_header( $columns ) {
-		$wc_actions = $columns['wc_actions'];
+		$wc_actions = isset( $columns['wc_actions'] ) ? $columns['wc_actions'] : '';
 		unset( $columns['wc_actions'] );
 
 		$columns['postnl_delivery_date'] = esc_html__( 'Delivery Date', 'postnl-for-woocommerce' );
--- a/src/Product/Single.php
+++ b/src/Product/Single.php
@@ -190,7 +190,7 @@ class Single {
 		$fields = self::product_field_maps( $this->service );
 
 		foreach ( $fields as $field ) {
-			if ( empty( $i ) && ! is_array( $_POST[ $field['id'] ] ) && 0 === $product->get_parent_id() ) {
+			if ( empty( $i ) && isset( $_POST[ $field['id'] ] ) && ! is_array( $_POST[ $field['id'] ] ) && 0 === $product->get_parent_id() ) {
 				$product->update_meta_data( $field['id'], sanitize_text_field( wp_unslash( $_POST[ $field['id'] ] ) ) );
 			} elseif ( ! empty( $i ) && 0 !== $product->get_parent_id() ) {
 				$field_value = ! empty( $_POST[ $field['id'] ][ $i ] ) ? $_POST[ $field['id'] ][ $i ] : '';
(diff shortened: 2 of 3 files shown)


src/Order/OrdersList.php
```diff
@@ -107,7 +107,7 @@ public function add_order_barcode_column_content( $column, $order_id ) {
 	 * @return array.
 	 */
 	public function add_order_delivery_date_column_header( $columns ) {
-		$wc_actions = $columns['wc_actions'];
+		$wc_actions = isset( $columns['wc_actions'] ) ? $columns['wc_actions'] : '';
```
Abdalsalaam, line 110: This will not fix the issue, if `wc_actions` is not exist, then the code will add empty value [here](https://github.com/Progressus-io/postnl-for-woocommerce/blob/6fd38f6fe7fcfcd6dc74c1dbe4c030acee40d4a1/src/Order/OrdersList.php#L114) 

But we should know why this warning appeared in the first place, this function is a call back for 
`manage_edit-shop_order_columns` and `manage_woocommerce_page_wc-orders_columns` filters and 'wc_actions' should be exist, if they are not ( removed by other plugin or any changes by 3rd parties ) I think we can just add the PostNL actions to the array without unset/set the wc-actions.

src/Product/Single.php
```diff
@@ -190,7 +190,7 @@ public function save_additional_product_shipping_options( $product_id, $i = '' )
 		$fields = self::product_field_maps( $this->service );
 
 		foreach ( $fields as $field ) {
-			if ( empty( $i ) && ! is_array( $_POST[ $field['id'] ] ) && 0 === $product->get_parent_id() ) {
+			if ( empty( $i ) && isset( $_POST[ $field['id'] ] ) && ! is_array( $_POST[ $field['id'] ] ) && 0 === $product->get_parent_id() ) {
```
Abdalsalaam, line 193: Lets simplify the code and return first if its not exist and also, if the key is not exist we should delete the meta data, for example :
the letterbox checkbox, if its checked the POST will contains a value 'YES' for the `_postnl_letterbox_parcel` key, but if its not checked it will not contains the key at all unless its a variations in this case it will be empty value, so this logic will just ignore it and the user will never be able to uncheck the field


```suggestion
			if ( ! isset( $_POST[ $field['id'] ] ) ) {
				$product->delete_meta_data( $field['id'] );
				continue;
			}

			if ( empty( $i ) && ! is_array( $_POST[ $field['id'] ] ) && 0 === $product->get_parent_id() ) {
```

Abdalsalaam (changes requested):

src/Order/OrdersList.php
```diff
@@ -107,12 +107,12 @@ public function add_order_barcode_column_content( $column, $order_id ) {
...
-		unset( $columns['wc_actions'] );
-
 		$columns['postnl_delivery_date'] = esc_html__( 'Delivery Date', 'postnl-for-woocommerce' );
-		$columns['wc_actions']           = $wc_actions;
-
+		if( isset( $columns['wc_actions'] ) ){
+			$wc_actions = $columns['wc_actions'];
+			unset( $columns['wc_actions'] );
+			$columns['wc_actions']       = $wc_actions;
+		}
```
Abdalsalaam, line 115: We are unsetting the **'wc_actions'** to add `postnl_delivery_date` before it, so this code will do nothing actually its just unset then set it.

Abdalsalaam (changes requested):

PR #188, opened 2024-09-18
Fix create return barcode when create return label is none

OlfatProgressus:
This is the fix of this issue 
https://github.com/Progressus-io/postnl-for-woocommerce/issues/187
task [link](https://app.clickup.com/t/8689ty0dy)

### DIFF
--- a/src/Order/Base.php
+++ b/src/Order/Base.php
@@ -509,9 +509,10 @@ abstract class Base {
 		$barcodes                        = $this->maybe_create_multi_barcodes( $label_post_data );
 		$label_post_data['main_barcode'] = $barcodes[0]; // for MainBarcode.
 		$label_post_data['barcodes']     = $barcodes;
+		$shipping_item_info 			 = new Shipping\Item_Info( $label_post_data );
 
-		$label_post_data['return_barcode']          = $this->maybe_create_return_barcode( $label_post_data );
-		$label_post_data['shipping_return_barcode'] = $this->maybe_create_shipping_return_barcode( $label_post_data, $label_post_data['main_barcode'] );
+		$label_post_data['return_barcode']          = $this->maybe_create_return_barcode( $label_post_data, $shipping_item_info );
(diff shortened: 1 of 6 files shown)


src/Order/Base.php
```diff
@@ -810,7 +810,13 @@ public function maybe_create_multi_barcodes( $post_data ) {
...
-		if ( 'shipping_return' !== $this->settings->get_return_shipment_and_labels() ||
+		$shipment_return_type = $this->settings->get_return_shipment_and_labels();
+
+		if ( 'none' === $shipment_return_type ) {
+			return '';
+		}
+
+		if ( 'shipping_return' !== $shipment_return_type ||
 		     'NL' !== $item_info->receiver['country'] ||
 		     '2928' === $item_info->shipment['shipping_product']['code'] ) {
```
Abdalsalaam, line 821: I think its better to -return first- on every case, like :
```php
		if ( 'none' === $shipment_return_type ) {
			return '';
		}

		if ( 'shipping_return' !== $shipment_return_type ) {
			return '';
		}

		if ( 'NL' !== $item_info->receiver['country'] ) {
			return '';
		}

		if ( '2928' === $item_info->shipment['shipping_product']['code'] ) {
			return '';
		}
```

src/Order/Base.php
```diff
@@ -810,7 +810,13 @@ public function maybe_create_multi_barcodes( $post_data ) {
 	 */
 	public function maybe_create_shipping_return_barcode( $post_data, $barcode ) {
 		$item_info = new Shipping\Item_Info( $post_data );
```
Abdalsalaam, line 812: I'm not sure what is the reason of this, and I think its should not be used like this, by this the code will create Item_Info object 3 times, once here, one within `maybe_create_return_barcode` function  and the latest in `create_label` function, this need to be fixed

Abdalsalaam (changes requested):

Abdalsalaam (approved):
I added some changes, thanks

PR #192, opened 2024-10-23
New: shipment addresses filter hook

alexmigf:
This pull request introduces a new filter `postnl_shipment_addresses` to modify shipment addresses before generating PostNL shipping labels. The filter is necessary to handle specific address formats used in our plugin, which appends the house number directly to the `address_1` field. Without this filter, the shipping label would display the house number twice—once in `address_1` and once in the `HouseNr` field.

### Usage:

```php
add_filter( 'postnl_shipment_addresses', function( $addresses, $client ) {
	if ( ! empty( $addresses[0]['HouseNr'] ) ) {
		$addresses[0]['HouseNr'] = '';
	}
	
	return $addresses;
}, 10, 2 );
```

### Benefits:

* Ensures correct address formatting for our plugin, preventing duplicated house numbers on PostNL shipping labels.
* Provides a flexible solution that could benefit other scenarios with different address formatting requirements.

### DIFF
--- a/src/Rest_API/Shipping/Client.php
+++ b/src/Rest_API/Shipping/Client.php
@@ -264,6 +264,6 @@ class Client extends Base {
 			);
 		}
 
-		return $addresses;
+		return apply_filters( 'postnl_shipment_addresses', $addresses, $this );
 	}
 }


Abdalsalaam:
Hello @alexmigf 
What is the issue you are trying to fix? we would be happy to help!

alexmigf:
Hi @Abdalsalaam 

I believe the reasoning is clearly outlined here:

> The filter is essential for handling specific address formats used by our plugin, where the house number is appended directly to the `address_1` field. Without this filter, the shipping label ends up displaying the house number twice—once in `address_1` and again in the `HouseNr` field.

Our goal is simply to gain more flexibility in formatting the address. By adding this filter, we can customize the address format to suit our needs and avoid duplicating the house number. This change has no impact on your existing code but provides a valuable customization option for our use case.

Abdalsalaam:
Hello @alexmigf,
Thanks for the reminder! I'm just trying to understand the issue and not create any other issues, because as I know it is a mandatory field for the API. Let me ask @JorisPostNL about this: will removing the `HouseNr` and adding it to `address_1` have any effects on your side?

alexmigf:
Hi @Abdalsalaam 

> Let me ask @JorisPostNL about this: will removing the HouseNr and adding it to address_1 have any effects on your side?

Based on our tests, everything functioned smoothly. We conducted testing with two individuals using different local setups.

Terminator-Barbapapa:
@Abdalsalaam @JorisPostNL 

In the mean time we also were in contact with Peter Koning at PostNL and came to the conclusion it might be better to remove the last occurrence of the house number (and suffix) from the `Street` value, instead of passing an empty house number value. 

As mentioned, passing an empty house number field does not interfere with the generation of a shipping label, but perhaps this causes issues later down the line when registering the shipment at PostNL. So the following example makes sure all fields have the correct values, even if the house number would also occur in the street name. Like for instance with `Plein 1944 19-A`

```
add_filter( 'postnl_shipment_addresses', function( $addresses, $client ) {

    if ( class_exists( 'WPO_WC_Postcode_Checker' ) ) {

        $street_name         = $addresses[0]['Street'];
        $house_number        = $addresses[0]['HouseNr'];
        $house_number_suffix = $addresses[0]['HouseNrExt'];

        if ( ! empty( $house_number ) ) {

            // Check if house number occurs in street name and get the position of the last occurrence
            $house_number_pos = strrpos( $street_name, $house_number );

            if ( $house_number_pos !== false ) {

                // Street name
                $addresses[0]['Street'] = rtrim( substr( $street_name, 0, $house_number_pos ) );

                // House number
                $addresses[0]['HouseNr'] = substr( $street_name, $house_number_pos, strlen( $house_number ) );

                // Suffix
                $addresses[0]['HouseNrExt'] = ltrim( substr( $street_name, $house_number_pos + strlen( $house_number ) ), '-' );

            }
        }
    }
    
    return $addresses;
}, 10, 2 );
```

In our testing we used multiple addresses including:

```
Plein 1944 19-A
6511 JC, Nijmegen 
```

Everything went with out any issues. As my colleague @alexmigf pointed out adding the filter will have no impact on your code base or functionality of the plugin.

JorisPostNL:
@alexmigf @Abdalsalaam @Terminator-Barbapapa If I'm not mistaken about your proposed change, the same effect can already be achieved by turning off both the address checker and Use PostNL Address Fields setting. When doing this and placing an order  with a difficult address, no splitting and mapping is being performed, with the complete address being sent to the API field Street, with the (mandatory) HouseNr field left empty:

"AddressType": "01",
					"CompanyName": "",
					"City": "Amsterdam",
					"Countrycode": "NL",
					"FirstName": "Test",
					"HouseNr": "",
					"HouseNrExt": "",
					"Name": "Home",
					"Street": "Plein '40-'45 2A",
					"Zipcode": "1063KP"
					
To be clear: an empty HouseNr field for domestic orders (NL>NL) is picked up by validation in the operation. The result of this is that an extra check is performed to determine the house number from the label which is then added to the HouseNr field associated with the shipment, which often comes with a surcharge for the customer, and in some cases it can cause other issues like specific productcodes reverting to base productcodes.
					
When placing an order for an "easy" address however, the splitting is done correctly: 

"AddressType": "01",
					"CompanyName": "",
					"City": "Den Haag",
					"Countrycode": "NL",
					"FirstName": "Test",
					"HouseNr": "3",
					"HouseNrExt": "",
					"Name": "Home",
					"Street": "Waldorpstraat",
					"Zipcode": "2521CA"

This is an indication complex splitting and mapping logic would have to be built. @Terminator-Barbapapa could you check if the filter you wrote is able to correctly split the address in the first AddressType 01 section I copied above?

Terminator-Barbapapa:
@JorisPostNL 

> If I'm not mistaken about your proposed change, the same effect can already be achieved by turning off both the address checker and Use PostNL Address Fields setting.

The goal here is to find a way for the PostNL plugin to play nicely with other plugins that might have been installed on the user's site. At the moment the PostNL plugin does not take into account that there could be other solutions running on a site that make changes to the checkout. 

Generating the shipping label with the proposed filter and the function I provided in my previous comment gives the following result:

```
array(1) {
  [0]=>
  array(10) {
    ["AddressType"]=>
    string(2) "01"
    ["CompanyName"]=>
    string(12) "The Company"
    ["City"]=>
    string(9) "AMSTERDAM"
    ["Countrycode"]=>
    string(2) "NL"
    ["FirstName"]=>
    string(4) "John"
    ["HouseNr"]=>
    string(1) "2"
    ["HouseNrExt"]=>
    string(1) "a"
    ["Name"]=>
    string(3) "Doe"
    ["Street"]=>
    string(13) "Plein '40-'45"
    ["Zipcode"]=>
    string(6) "1063KP"
  }
}
```

So this runs without any issues, and even makes sure all fields have their corresponding values, including house number and house number extension.

JorisPostNL:
Thank you for testing @Terminator-Barbapapa. I have made note to discuss this proposal during our next development meeting and will get back to you about this asa

Terminator-Barbapapa:
Thank you @JorisPostNL. At the moment this is holding back our release of the integration with the new PostNL adres autocomplete. So it would be in both of our interest to solve this issue as soon as possible.

Abdalsalaam:
Hello @alexmigf @Terminator-Barbapapa 
Thanks for this contribution, this will be included in version 5.6.5, which will be released soon.

Abdalsalaam:
Hello @alexmigf, Please note that this PR is released to public within v5.7.0
thanks!

alexmigf:
@Abdalsalaam We appreciate you letting us know.

PR #193, opened 2024-10-24
Id check with pickup points

OlfatProgressus:
This PR for these tasks 
1. [Add ID check shipping option to Pick-up points](https://app.clickup.com/t/8689buv73)
2. [Add product ID Check + Insured Shipping for domestic](https://app.clickup.com/t/8685026vy)
3. [Add pickup as a new zone to default shipping options](https://app.clickup.com/t/8689buv73?comment=90110084820548&threadedComment=90110089618195)

### DIFF
--- a/src/Order/Bulk.php
+++ b/src/Order/Bulk.php
@@ -130,8 +130,7 @@ class Bulk extends Base {
 	 * @return string
 	 */
 	public function bulk_action_change_shipping_options( $redirect, $doaction, $object_ids ) {
-
-		if ( 'postnl-change-shipping-options' !== $doaction ) {
+		if ( 'postnl-change-shipping-options' !== $doaction || empty( $object_ids ) ) {
 			return $redirect;
 		}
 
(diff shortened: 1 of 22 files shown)


src/Order/Bulk.php
```diff
@@ -162,7 +162,12 @@ public function bulk_action_change_shipping_options( $redirect, $doaction, $obje
 					);
 				}
 
-				if ( ! $have_label_file && $match_shipping_zones ) {
+				if ( ! $have_label_file && 'PICKUP' === $zone ) {
+					$meta = $order->get_meta( $this->meta_name );
+					$meta['backend'] = $selected_shipping_options ;
+					$order->update_meta_data( $this->meta_name, $meta );
+					$order->save();
+				}elseif ( ! $have_label_file && $match_shipping_zones ) {
```
Abdalsalaam, line 170: I don't understand the purpose of this change

`! $have_label_file && 'PICKUP' === $zone` and `! $have_label_file && $match_shipping_zones` will be always the same, due to this change :
`$match_shipping_zones = $zone === $this->get_shipping_zone( $order ) || 'PICKUP' === $zone;`

Abdalsalaam (changes requested):

src/Order/Bulk.php
OlfatProgressus: @Abdalsalaam 
```
if ( ! $have_label_file && 'PICKUP' === $zone ) {
					$meta = $order->get_meta( $this->meta_name );
					$meta['backend'] = $selected_shipping_options ;
					$order->update_meta_data( $this->meta_name, $meta );
					$order->save();
				}
```
				this will be applied if the order has drop-off option from frontend to make sure it will not remove the frontend data

PR #194, opened 2024-11-04
update WP version and fix PCP errors

OlfatProgressus:


templates/checkout/postnl-delivery-day.php
```diff
@@ -53,7 +53,7 @@ class="<?php echo esc_attr( $option['type'] . ' ' . $is_active ); ?>"
 									name="<?php echo esc_attr( $data['field_name'] ); ?>" 
 									class="postnl_sub_radio" 
 									value="<?php echo esc_attr( $value ); ?>"
-									<?php echo $is_checked; ?>
+									<?php echo wp_kses_post( $is_checked ); ?>
 								/>
```
Abdalsalaam, line 57: $is_checked is boolean, so no need for this

templates/checkout/postnl-delivery-day.php
```diff
@@ -53,7 +53,7 @@ class="<?php echo esc_attr( $option['type'] . ' ' . $is_active ); ?>"
 									name="<?php echo esc_attr( $data['field_name'] ); ?>" 
 									class="postnl_sub_radio" 
 									value="<?php echo esc_attr( $value ); ?>"
-									<?php echo $is_checked; ?>
+									<?php echo wp_kses_post( $is_checked ); ?>
```
Abdalsalaam, line 56: I think no need for this

readme.txt
```diff
@@ -3,7 +3,7 @@ Contributors: PostNL, shadim, abdalsalaam
 Tags: woocommerce, export, delivery, packages, PostNL, Shipping
 Requires at least: 4.6
 Requires PHP: 5.6
-Tested up to: 6.6
+Tested up to: 6.7
 Stable tag: 5.6.2
 WC requires at least: 4.0
 WC tested up to: 9.3
```
Abdalsalaam, line 9: please bump this to 9.4 since its released and we already tested it.

src/Rest_API/Shipping/Item_Info.php
```diff
@@ -993,7 +993,8 @@ protected function check_insurance_amount_limit( $backend_data, $order_total ) {
 		} // For EU shipments, validate that insurance does not exceed €5000
 		elseif ( ! $is_non_eu_shipment && 'yes' === $backend_data['insured_shipping'] && $order_total > 5000 ) {
 			throw new \Exception(
-				__( 'Insurance amount for EU shipments cannot exceed €5000. Your total is: ' . $order_total, 'postnl-for-woocommerce' )
+				// Translators: %s is the order total.
+				sprintf( __( 'Insurance amount for EU shipments cannot exceed €5000. Your total is: %s', 'postnl-for-woocommerce' ), wp_kses_post( $order_total ) )
```
Abdalsalaam, line 997:
```suggestion
				sprintf( esc_html__( 'Insurance amount for EU shipments cannot exceed €5000. Your total is: %d', 'postnl-for-woocommerce' ), $order_total )
```
$order_total is a number so we can use %d instead of %s

src/Utils.php
```diff
@@ -692,8 +692,9 @@ public static function is_eligible_auto_letterbox( $order ) {
 			$products = $order->get_cart();
 		}
 
-		$is_eligible = self::check_products_for_letterbox( $products );
-
+		if( isset( $products ) ) {
+			$is_eligible = self::check_products_for_letterbox( $products );
+		}
```
Abdalsalaam, line 697: Please extract cart logic from this function, you can create a new function like :

```php
	public static function is_cart_eligible_auto_letterbox( $cart ) {
		if ( ! in_array( WC()->customer->get_shipping_country(), Utils::get_available_country_for_letterbox(), true ) ) {
			return false;
		}

		return self::check_products_for_letterbox( $cart->get_cart() );
	}
```

and `is_order_eligible_auto_letterbox` function will be like :
```php
	public static function is_order_eligible_auto_letterbox( $order ) {
		if ( wc_get_base_location()['country'] == 'BE' ) {
			return false;
		}

		// Check if order id provided.
		if ( is_int( $order ) ) {
			$order = wc_get_order( $order );
		}

		if ( ! is_a( $order, 'WC_Order' ) ) {
			return false;
		}

		if ( $order->meta_exists( '_postnl_letterbox' ) ) {
			return (bool) $order->get_meta( '_postnl_letterbox', true );
		}

		if ( ! in_array( $order->get_shipping_country(), Utils::get_available_country_for_letterbox(), true ) ) {
			$order->update_meta_data( '_postnl_letterbox', false );
			$order->save_meta_data();

			return false;
		}

		$products    = $order->get_items();
		$is_eligible = self::check_products_for_letterbox( $products );

		$order->update_meta_data( '_postnl_letterbox', $is_eligible );
		$order->save_meta_data();

		return $is_eligible;
	}
```

Abdalsalaam (changes requested):

src/Utils.php
OlfatProgressus: @Abdalsalaam updated

readme.txt
OlfatProgressus: @Abdalsalaam updated

templates/checkout/postnl-delivery-day.php
OlfatProgressus: @Abdalsalaam Removed

PR #195, opened 2024-11-11
not to remove the merged file path from the saved data

OlfatProgressus:
This PR contain fix for[ this issue ](https://app.clickup.com/t/868apk1qn)
1. Remove the code that unset merged_files from saved_data array
2. Remove the code that delete the files after merge

### DIFF
--- a/src/Order/Base.php
+++ b/src/Order/Base.php
@@ -537,14 +537,7 @@ abstract class Base {
 			$barcodes
 		);
 
-		$saved_data['labels'] = array_map(
-			function ( $label ) {
-				unset( $label['merged_files'] );
-
-				return $label;
-			},
-			$labels
-		);
+		$saved_data['labels'] = $labels;
 
 		if ( $this->settings->is_auto_complete_order_enabled() ) {
 			// Updating the order status to completed.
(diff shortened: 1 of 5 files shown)


src/Order/Base.php
```diff
@@ -1277,6 +1271,12 @@ public function delete_label( $saved_data ) {
 			return false;
 		}
 
+		if( isset( $saved_data['labels']['label']['merged_files'] ) ){
+			foreach ( $saved_data['labels']['label']['merged_files'] as  $label_path ) {
+				unlink( $label_path );
+			}
+		}
```
Abdalsalaam, line 1278:
```suggestion
		if ( isset( $saved_data['labels']['label']['merged_files'] ) ) {
			foreach ( $saved_data['labels']['label']['merged_files'] as $label_path ) {
				unlink( $label_path );
			}
		}
```

src/Order/Base.php
```diff
@@ -711,7 +704,8 @@ public function put_label_content( $response, $order, $parent_barcode, $parent_l
 					$label_extension = ! empty( $label_contents['OutputType'] ) ? sanitize_title( $label_contents['OutputType'] ) : 'pdf';
 					$barcode         = $response[ $type ][ $shipment_idx ][ $content_type['barcode_key'] ];
 					$barcode         = is_array( $barcode ) ? array_shift( $barcode ) : $barcode;
-					$filename        = Utils::generate_label_name( $order->get_id(), $label_type, $barcode, 'A6', $label_extension );
+					$label_format  	 = $this->settings->get_label_format();
+					$filename        = Utils::generate_label_name( $order->get_id(), $label_type, $barcode, $label_format, $label_extension );
```
Abdalsalaam, line 708: but the label file returned from API is always **A6**, is there a reason for this?

Abdalsalaam (changes requested):

src/Order/Base.php
OlfatProgressus: @Abdalsalaam It cause a fatal error when it generate files with format A4 then try to create the label again from bulk action because file path is A6 but the one saved to database is A4

Abdalsalaam (approved):

PR #205, opened 2025-01-24
Add as soon as possible option to delivery options tab

OlfatProgressus:
### All Submissions:

Closes # .
1. [Task url](https://app.clickup.com/t/868bqw4a0)
2. [Task url](https://app.clickup.com/t/868bp829x)

1. Disable Delivery day option
2. Open checkout
3. Delivery days tab will be shown with as soon as possible option that has the first option values
4. Place an order
5. Create label for this order

### Other information:

### Changelog entry


src/Frontend/Base.php
```diff
@@ -100,9 +100,6 @@ abstract public function is_enabled();
 	 * Collection of hooks when initiation.
 	 */
 	public function init_hooks() {
-		if ( ! $this->is_enabled() ) {
-			return;
-		}
```
Abdalsalaam, line 103: Instead of removing this, I suggest to modify `is_enabled` function to let it return `true`, and create a new function to check if days options should be displayed for the customer or not, for example : 
```php
	/**
	 * Check if this feature is enabled from the settings.
	 *
	 * @return bool
	 */
	public function is_enabled() {
		return true;
	}

	/**
	 * Check if customer allowed to choose a delivery day.
	 *
	 * @return bool
	 */
	public function is_customer_allowed_to_pick_delivery_day() {
		return $this->settings->is_delivery_days_enabled();
	}
```

Abdalsalaam (changes requested):

src/Frontend/Base.php
OlfatProgressus: @Abdalsalaam I pushed the changes please review

Abdalsalaam (approved):
LGTM! thanks

PR #206, opened 2025-01-31
change shipping zone for TF and GC states to be ROW instead of EU

OlfatProgressus:
### All Submissions:

Closes # .
Task [link](https://app.clickup.com/t/8685xyjbb)

1. Create order with shipping address country "Spain" and province "Santa Cruz de Tenerife"
2. Check Delivery type from L&T tab , it should be Non EU 
3. Create label and check code , it should be 4909

### Other information:

### Changelog entry

### DIFF
--- a/src/Order/Base.php
+++ b/src/Order/Base.php
@@ -121,7 +121,7 @@ abstract class Base {
 		}
 
 		// Get from the plugin settings
-		$delivery_zone = $this->get_shipping_zone( $order );
+		$delivery_zone = Utils::get_shipping_zone( $order->get_shipping_country(), $order->get_shipping_state() );
 		$frontend_data = $this->get_frontend_data( $order->get_id() );
 
 		if ( ! empty( $frontend_data['dropoff_points'] ) ) {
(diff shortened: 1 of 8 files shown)


src/Order/Base.php
```diff
@@ -144,11 +144,16 @@ public function get_shipping_options( $order ) {
...
 	public function get_shipping_zone( $order ) {
 		$shipping_destination = $order->get_shipping_country();
+		$shipping_state       = $order->get_shipping_state();
 
 		if ( in_array( $shipping_destination, array( 'NL', 'BE' ) ) ) {
 			return $shipping_destination;
 		}
 
+		if ( in_array( $shipping_state, array( 'TF', 'GC' ) ) ) {
+			return 'ROW';
```
Abdalsalaam, line 154: `Grand Cape Mount` in Liberia have the same state code `GC`, we should check if the country is Spain as well.

Abdalsalaam (approved):
LGTM, I pushed some changes to remove code duplication and fix Spanish states check
thank you

PR #251, opened 2025-10-06
Remove the non working dublicated container block

OlfatProgressus:
### All Submissions:

* [ ] Does your code follow [WooCommerce](https://docs.woocommerce.com/document/create-a-plugin/) and [WordPress](https://make.wordpress.org/core/handbook/best-practices/coding-standards/) standards?

Closes # .

1.
2.
3.

### Other information:

### Changelog entry

### DIFF
--- a/assets/css/fe-checkout.css
+++ b/assets/css/fe-checkout.css
@@ -264,3 +264,7 @@ tr.postnl-co-tr-container > td {
 .postnl-hidden{
 	visibility: hidden !important;
 }
+
+.wc-block-components-checkout-order-summary__content#\:r2\:  .postnl_checkout_container {
+	display: none;
+}


assets/js/fe-checkout.js
```diff
@@ -1,4 +1,25 @@
...
+	 */
+	document.addEventListener( 'DOMContentLoaded', () => {
+		const observer = new MutationObserver( () => {
+			// Only run on mobile screens
+			const containers = document.querySelectorAll(
+				'.is-mobile .wc-block-checkout__sidebar .postnl_checkout_container'
+			);
+
+			if ( containers.length > 0 ) {
+				containers.forEach( el => ( el.style.display = 'none' ) );
```
Abdalsalaam: It still isn’t working as expected. The reason is that the IDs are duplicated. When the elements’ IDs are duplicated on the page, the JS won’t work properly. **The IDs should be unique**.

OlfatProgressus: Unique IDs added, I tested it from my side and it is working

PR #304, opened 2026-03-11
Automatic printing

OlfatProgressus:
### All Submissions:

Closes # .
[Task url](https://app.clickup.com/t/861myrbhf)

1. Create an order, open the order edit page, and generate the label. You should now see three buttons: Print, Download, and Delete.
2. Go to the Orders listing page and generate a label using Bulk Actions. After the label is created, a Print Label link should appear in the success message.
3. In the Actions column, verify that a new printer icon has been added for the Print action.

### Other information:

### Changelog entry


Abdalsalaam:
## 🤖 Automated PR review
**ClickUp task:** [automatic printing (parent ticket)](https://app.clickup.com/t/861myrbhf)
**Verdict:** 🟡 Needs changes

> ⚠️ **Process flag:** This PR overlaps **#307** ("Add auto-create label on Print/Download for single order page", subtask `868j7zh6r` of this same parent ticket). Both edit `src/Order/Single.php`, `src/Order/Bulk.php`, and the two admin CSS files, and both introduce a print path — but with **divergent mechanisms** (this PR rasterizes via pdf.js → print window; #307 opens the print window synchronously and adds the bulk Print/Download AJAX flow, the `admin-order-bulk.js` handler, a changelog and README entry). Merging both as-is will produce conflicting button markup and a double-implemented print path. **Please decide explicitly which PR is the source of truth before either merges** and close/rebase the other. #307 appears to be the more complete implementation of the print flow.

**Spec coverage**
- ✅ **Single order page: Print / Download / Delete buttons** — `Single.php` `meta_box_html` now renders a `button-print-label`, a Download `<a>`, and a Delete `<a>`.
- ✅ **Bulk success message: "Print Label" link** — added in `Bulk.php` `merge_bulk_labels` and `generate_label_and_notes`.
- ✅ **Orders-list Actions column: printer icon** — `postnl-print-label` action in `add_create_label_actions_button` + SVG in `admin-order-bulk.css`.
- ⚠️ **Core "send straight to printer" intent** — a print path exists but is manually triggered and (see below) is likely blocked by popup blockers; auto-trigger-on-creation lives in #307, not here.
- ❌ **Popup-blocker handling (reviewer-requested)** — `window.open` runs after async render, handle is not null-checked, blocked popups only `console.error`.
- ✅ **Nonce / capability on new AJAX (reviewer-requested)** — no new AJAX added; print/download reuse `get_label_file`/`get_bulk_file`, which keep `check_ajax_referer` + `current_user_can('manage_woocommerce'/'view_order')`. Intact.

**Findings**
- 🟠 **[High] popup-blocker / logic** — `window.open('','_blank')` is called inside the post-render `.then()` callback (`admin-print-label.js:50`), not synchronously in the click handler (`:72`). Browsers block `window.open` outside a direct user-gesture stack, so both the single-order Print button and the bulk "print label" link will be blocked under default settings — fix: open the window synchronously in the click handler, keep the reference, then write into it after the PDF renders (the pattern #307 already uses).
- 🟠 **[High] null deref** — the `window.open` result is dereferenced immediately (`printWindow.document.open()/write()/close()`, `:51-53`) with no null guard; when the popup is blocked this throws and is swallowed by `.catch` → button silently does nothing — fix: `if ( ! printWindow ) { /* surface "allow popups" / fall back to download */ return; }`.
- 🟠 **[High] WP.org guideline / supply-chain** — pdf.js is loaded from `cdnjs.cloudflare.com` (`pdf.min.js` + `pdf.worker.min.js`) in both `Single.php` and `Bulk.php`, registered with a `null` version and no SRI. WP.org disallows loading executable JS from remote CDNs; it also breaks offline/restricted environments — fix: vendor both files into `assets/js/` and enqueue locally with `POSTNL_WC_VERSION`, pointing `workerSrc` at the local copy.
- 🟡 **[Medium] print may not fire / blank pages** — `printWindow.onload` is assigned *after* the synchronous `document.write()`/`document.close()` (`:51-57`), and the body is `` rendered into `about:blank`. The load event may have already fired (or never fires), and data-URL images may not be decoded before `print()`, so the dialog can silently fail or print blank — fix: inject an inline `<script>window.onload=...print()</script>` into the written HTML, or `Promise.all` over `img.decode()` before calling print.
- 🟡 **[Medium] missing user-facing error handling** — every failure path (fetch fail, render fail, blocked popup) ends in a bare `console.error` (`:58`). The single-order box has `#shipment-postnl-error-text` and bulk uses admin notices, but neither is used — fix: surface a visible error (reuse `#shipment-postnl-error-text` on the single page; notice/alert on the list) and specifically detect the blocked-popup case.
- 🟡 **[Medium] missing `pdfjsLib` guard** — `postnlPrintLabel()` calls `pdfjsLib.getDocument(...)` with no existence check; if the CDN script fails to load, every print click throws `ReferenceError` with no degradation (`:13`) — fix: guard `typeof pdfjsLib` and fall back to the download URL when missing.
- 🟢 **[Low] DRY / duplicate enqueue** — the ~25-line `postnl-pdfjs` + `postnl-admin-print-label` + `postnlPrintLabelData` block is duplicated verbatim in `Single.php::enqueue_order_single_css_script` and `Bulk.php::enqueue_bulk_assets` (and double-registers on the HPOS orders screen) — fix: extract one shared helper (e.g. in `Order\Base`) so the CDN-removal/version fix lives in one place.
- 🟢 **[Low] i18n** — the print-window `<title>Print Label</title>` (`admin-print-label.js:43`, visible as the tab/print header) is hardcoded English — fix: pass it through the existing `postnlPrintLabelData` localize payload (e.g. `printTitle => esc_html__( 'Print Label', ... )`).
- 🟢 **[Low] performance** — every page is rasterized at `getViewport({scale:2})` to a PNG data URL and concatenated into one HTML string (`admin-print-label.js`); large bulk batches can spike memory with no bound — fix: lower/auto scale, or print the PDF directly via an embedded object/iframe rather than rasterizing every page.
- ℹ️ **[Info] no changelog / version bump** — this is a user-facing feature but the diff has no `readme.txt`/`changelog.txt`/version change (#307 does); add per the release checklist. No automated tests were added despite the existing `tests/js/` Jest harness.

**Testing notes**
1. With a popup blocker at default settings (Chrome/Firefox/Safari), click **Print Label** on the single-order box and the bulk "print label" link — confirm whether the print window actually opens (expect it to be blocked with current async `window.open`).
2. Block/disable network access to `cdnjs.cloudflare.com` and click Print — confirm graceful behavior vs a silent `ReferenceError`.
3. Print a multi-page / bulk-merged label and verify all pages render (not blank/partial) and page breaks are correct.
4. Confirm the orders-list printer icon and the relabeled "PostNL Download Label" action both still hit the existing nonce-protected `get_label_file`/`get_bulk_file` handlers (admin-only; not a checkout change, so no classic-vs-blocks impact).
5. Coordinate with the #307 assignee to verify the two PRs don't both enqueue `postnl-pdfjs` / diverge on the print mechanism after merge.

PR #305, opened 2026-03-12
Fix ALA detection for NL and international Shipments

OlfatProgressus:
### All Submissions:

Closes # .
[Task url](https://app.clickup.com/t/868hv048h)

1. Place order with Belgium address and eligible for letterbox product
2. Add same product to cart and go to the checkout page again
3. change the country to Netherland, you shouldn’t see the delivery days menu now, if the address is correct you should see "These items are eligible for letterbox delivery."
4. when back to Belgium, you should see the delivery days menu

### Other information:

### Changelog entry


client/checkout/postnl-container/block.js
```diff
@@ -56,7 +56,9 @@ export const Block = ( { checkoutExtensionData } ) => {
 	const { setExtensionData } = checkoutExtensionData;
 	const postnlData = getSetting( 'postnl-for-woocommerce-blocks_data', {} );
 
-	const letterbox = postnlData.letterbox || false;
+	const [ isLetterbox, setIsLetterbox ] = useState(
+		postnlData.letterbox || false
```
Abdalsalaam, line 99: I think adding `is_letterbox` to the early-exit responses (lines 343-351, 365-373 of `Extend_Block_Core.php`) is better than this, so the frontend always gets an explicit letterbox signal

Abdalsalaam (commented):
LGTM, can you check the comment I added? and also please add a changelog
thanks

PR #309, opened 2026-04-21
Add secondary API key

sal4sup:
### All Submissions:

Closes # .
https://app.clickup.com/t/868j4b857

1.Go to Postnl settings
2. Validate the new API key 
3.

### Other information:

### Changelog entry


Abdalsalaam:
## 🤖 Automated PR review
**ClickUp task:** [Second API Key field requirements](https://app.clickup.com/t/868j4b857)
**Verdict:** 🟡 Needs changes

This PR is more complete than its terse title suggests — all four requirement groups are present. The blocker is a save-time correctness bug that can silently defeat the key switch, plus a NewKey-header semantics mismatch against the spec.

**Spec coverage**
- ✅ **1. "New API Key" field + ENG/NL tooltip** — `api_keys_new` added in `Settings.php` with the exact English description; `nl_NL.po` carries the exact NL strings.
- ✅ **2a. Dismissible banner in config + order overview** — `Api_Key_Banner::is_target_screen()` covers `wc-settings`+PostNL section, `edit-shop_order`, and `wc-orders` (HPOS). Rendered as a top-of-page `admin_notices` warning, not embedded inside the section (see Low finding — confirm with requester).
- ✅ **2b. "Remind me later" + permanent dismiss** — remind meta cleared on `wp_login`; dismiss is permanent per user.
- ⚠️ **3a. Switch original → new key when valid** — `get_effective_api_key()` gates correctly, but the singleton-staleness bug (High) can store the validated hash over the wrong key and silently keep the old key.
- ✅ **3b. Test call on save (Barcode API)** — `process_new_api_key_validation()` → `Key_Validator::validate()` does a live `v1_1/barcode` GET. (Production-only by design — see Low.)
- ✅ **3c. Exact ENG/NL error + don't use field unless validated** — exact copy via `WC_Admin_Settings::add_error`; `set_api_key_new_validated(false)` falls back to original.
- ⚠️ **4. NewKey header No/Yes/Same on every call** — injected in `Base::get_basic_headers_args()` (covers all keyed clients) and the validator. But `Yes` is reported for any distinct-but-unvalidated key, diverging from the spec's "entered, **works**, different" (see Medium).
- ✅ **Security: key never exposed client-side** — banner localize passes only `ajaxUrl`+`action`; JS posts only `action`/`nonce`/`mode`; AJAX handler is admin-only (`manage_woocommerce` + nonce). Key stays server-side.

**Findings**
- 🟠 **[High] correctness/race** — `set_api_key_new_validated(true)` re-reads the key from the `Settings` singleton (`get_api_key_new()` → `get_option()`), but `parent::process_admin_options()` updated the DB and the **shipping-method** object, not the singleton. The singleton is instantiated earlier in the request (e.g. `Main::get_logger()` at `Main.php:507`, banner `should_show()`) and caches `$this->settings` in memory, so the SHA-256 hash is computed over the stale/old key. On the next request `get_effective_api_key()` hashes the new key, `hash_equals` fails, and the validated flag is silently lost — the switch never happens. `src/Shipping_Method/PostNL.php:103-126` + `src/Shipping_Method/Settings.php:855-883` — **fix:** hash the exact key just validated (e.g. `set_api_key_new_validated( true, $new_key )`) or call `init_settings()` on the singleton before reading.
- 🟡 **[Medium] spec-compliance** — `get_new_key_header_value()` returns `Yes` as soon as a distinct key is typed, even if validation failed (the doc comment states this intentionally). PostNL then reads `NewKey: Yes` ("works") while traffic is still authenticated with the original key. Spec #4 defines `Yes` = "entered, **works**, different". `src/Shipping_Method/Settings.php:925-942` — **fix:** gate `Yes` on `is_api_key_new_validated()`; confirm the desired value for the entered-but-unvalidated state with the task owner.
- 🟢 **[Low] i18n/WPCS** — translatable string with a `%d` placeholder lacks the required `// translators:` comment (triggers `WordPress.WP.I18n.MissingTranslatorsComment`). `src/Rest_API/Barcode/Key_Validator.php` (the `Unexpected HTTP status %d` `sprintf`) — **fix:** add `// translators: %d is the HTTP status code returned by PostNL.` on the line directly above.
- 🟢 **[Low] lifecycle** — new persistent data has no cleanup: option `postnl_api_keys_new_validated_hash` and per-user meta `postnl_new_api_key_banner_dismissed` / `_remind_later`. `uninstall.php` (not modified) — **fix:** add `delete_option` + per-user `delete_metadata` for the new keys.
- 🟢 **[Low] UX/AJAX-resilience** — `new-api-key-banner.js` removes the banner in `.always()`, so a failed dismiss (expired nonce / 403) still hides it for this view while server meta is unset → it reappears next load. `assets/js/new-api-key-banner.js` — **fix:** only remove on a `success` response; otherwise show an inline error / leave it for retry.
- 🟢 **[Low] UX/sandbox** — validation and the key swap are production-only (`process_new_api_key_validation()` early-returns when `environment_mode !== 'production'`; `get_effective_api_key()` is bypassed for sandbox in `Base::set_api_key()`). A new key entered in sandbox is saved but never validated or used. `src/Shipping_Method/PostNL.php:84-88`, `src/Rest_API/Base.php:157-162` — **fix:** confirm production-only is intended (matches migration intent); the spec does not restrict validation to production.
- 🟢 **[Low] spec-interpretation** — banner is a standard top-of-page `admin_notices` warning, not a banner embedded "in the config section / order overview section". Functionally present on both screens. `src/Admin/Api_Key_Banner.php` (`maybe_render`) — **fix:** confirm a top-of-screen notice satisfies "in the section"; if embedded is required, hook the settings section / orders header.
- ℹ️ **[Info] consistency** — `api_keys_new` and `NEW_API_KEY_VALIDATED_HASH_OPTION` are global, while original keys are country-scoped (`get_api_key()` via `get_country_option`). On a base-country change (NL↔BE) the new key + validated hash carry over against a now-different original. Latent (single base country per site). `src/Shipping_Method/Settings.php` — **fix:** mirror country scoping or document the new key as intentionally global.

**Testing notes**
- Save settings in production with a valid distinct new key, then trigger any outbound call (generate a barcode/label) on a **second** request and confirm it authenticates with the new key — this is the path the High staleness bug breaks.
- Enter an invalid distinct key: confirm the exact ENG/NL error shows, the plugin keeps using the original key, and check the `NewKey` header value sent (currently `Yes` — verify against the spec's intended value for unvalidated keys).
- Verify the banner on both screens (PostNL settings section and the orders list, including HPOS `wc-orders`); test "Remind me later" returns after logout/login and "Dismiss permanently" stays gone; force a 403 (expired nonce) on dismiss and confirm desired behavior.
- This change is admin/settings + header injection only — no classic-vs-blocks checkout rendering or fee logic is touched, so checkout behavior should be symmetric; a quick smoke test of one labeled order in each mode is still worthwhile.
- Regenerate `.pot` (`npm run makepot` / `updatepo`) so the new English strings get canonical source refs and verify msgids match the spec verbatim.

dustinparker:
I went through this against `main` and the PostNL requirements in [868j4b857](https://app.clickup.com/t/868j4b857). Seven things to sort before it merges. Items 1 through 6 all sit in the same area of the code, so they are probably one pass rather than six separate ones.

Worth saying up front: merging this PR is what makes `Service_Factory::has_v4_key()` live. It currently wraps `get_api_key_new()` and `is_api_key_new_validated()` in `method_exists()` guards that return false, because those accessors only exist on this branch. The moment this lands, the V4 gate becomes reachable. The flags still default off so nothing routes without a filter, but it means the key handling here needs to be right before merge rather than after.

### 1. The validation call tests the wrong API

`src/Rest_API/Barcode/Key_Validator.php:52`

```php
$endpoint = POSTNL_WC_PROD_API_URL . '/shipment/v1_1/barcode';
```

That is the same endpoint as `src/Rest_API/Legacy/Barcode/Client.php:29`, so it only proves the key works against V1. The whole point of this field is to confirm the merchant holds a V4-capable key, and this call cannot tell the difference.

It matters because `Service_Factory::has_v4_key()` gates all V4 routing on `is_api_key_new_validated()`, which this sets. A merchant with a non-V4 key enters it here, passes validation against V1, and every flow with its flag on then routes to V4 where the key has no access. They would be better off leaving the field empty.

I think the validation call should go through the SDK against a V4 endpoint. That is the only thing that actually proves V4 access.

### 2. The endpoint is hardcoded to production

Same line. `AGENTS.md` is explicit about this: "Always use `$settings->is_sandbox()` to determine which URL to use. Never hardcode `https://api.postnl.nl`." Moving the call to the SDK fixes this at the same time, since `Client_Factory::build()` already takes `$is_sandbox`.

### 3. There is no sandbox key field

Only `api_keys_new` exists. PostNL issued four keys, two sandbox and two production, because they are environment-scoped in the same way the existing ones are.

`src/Rest_API/Base.php:157`

```php
if ( true === $this->is_sandbox ) {
	$this->api_key = $this->settings->get_api_key_sandbox();
	return;
}

$this->api_key = $this->settings->get_effective_api_key();
```

Sandbox never reaches the new-key logic. We need `api_keys_sandbox_new` following the same pattern as `api_keys` and `api_keys_sandbox`, plus the matching show/hide line in `admin-settings.js`.

### 4. Validation cannot run in sandbox

`src/Shipping_Method/PostNL.php:92`

```php
$env = $this->get_option( 'environment_mode' );
if ( 'production' !== $env ) {
	return;
}
```

So `is_api_key_new_validated()` never gets set in sandbox, `has_v4_key()` is always false there, and the V4 path cannot be exercised in sandbox at all. That blocks [subtask 28](https://app.clickup.com/t/868jqaqj3), the flag flips. Once the validator is environment-aware this early return can go.

### 5. The validated flag is not environment-scoped

`NEW_API_KEY_VALIDATED_HASH_OPTION` is a plain option with no environment in the key. Validate in production, switch to sandbox, and the flag stays set. `has_v4_key()` then returns true and V4 routes with a production key against the sandbox host, which fails in a way that looks like a V4 bug rather than a config one. Either the option name or the hashed value needs the environment in it.

### 6. The NewKey header reports "Yes" before validation

`Settings::get_new_key_header_value()` only compares empty and equal, so a key that failed validation still reports `Yes`. The merchant sees the invalid key error while every call tells PostNL they have adopted.

The requirements define `Yes` as "this means the secondary API key field is entered, works and is different from the original key". Works means validated, so I think this should gate on `is_api_key_new_validated()`. That also settles the open question in the docblock on `test_distinct_new_key_reports_yes_without_validation`, which is waiting on exactly this.

One decision while you are in there: if an entered key has not validated yet, `No` understates it and there is no fourth value in the spec. I will confirm with PostNL which they want.

Separately, the header goes out on every call from `Base::get_basic_headers_args()` including sandbox, where it reports the production field's value. It should follow the environment once item 3 is in.

### 7. The branch is 12 commits behind main

Would you be able to merge `main` in and push? CI last ran on 27 July, before all twelve of those commits, so the green check on this PR is not telling us anything about the current state.

The merge itself is clean, I tested it locally. One thing to watch: `tests/php/bootstrap-unit.php` is the only file both sides touch, and it auto-merges into two `WC_Settings_API` stubs. Main's is deliberately empty so that a test double reaching a real parent method fatals loudly instead of quietly returning stub data. This branch's is functional with an in-memory store. Both are wrapped in `class_exists()`, so main's wins and this one becomes unreachable. Nothing breaks today because `SettingsNewApiKeyTest` uses partial mocks and never reaches the parent, but please drop this branch's stub and keep main's.

---

Joris's read-only request from this morning is deliberately not on this list. That one is still open with him and I would rather not hold this PR for it.

sal4sup:
Thanks for the really thorough pass, @dustinparker — this was exactly the kind of review that catches the things that bite later. I've worked through all seven and merged `main` in. Here's where each one landed:

**1 & 2 — validation now actually proves V4 access.** You were right that hitting the V1 barcode endpoint told us nothing useful. `Key_Validator` goes through the SDK now — `Client_Factory` → `barcodes()->generateBarcode()` (`POST /shipment/delivery/v4/barcode`) — and the environment comes from the caller instead of a hardcoded host. Auth failures come back as an invalid-key error, retryable/exhausted ones are treated as transient so we don't disable a good key on a network blip, and everything else is surfaced through `Exception_Converter`.

**3 — sandbox has its own new-key field now.** Added `api_keys_sandbox_new` (+ `get_api_key_sandbox_new()`), and `get_api_key_new()` is environment-aware so every V4 consumer just gets the right key. `Base::set_api_key()` no longer needs its sandbox branch — it calls `get_effective_api_key()`, which resolves the validated new key per environment and falls back to the legacy key. Matching show/hide is in `admin-settings.js`.

**4 — validation runs in sandbox too.** Dropped the `'production' !== $env` early return, so the sandbox V4 path can finally be exercised.

**5 — the validated flag is per-environment.** It gets a `_sandbox` suffix now (`validated_hash_option_name()`), so validating one environment can't accidentally vouch for the other.

**6 — NewKey is gated on validation and follows the environment.** `Yes` only when the distinct key has actually passed validation — I flipped the `test_distinct_new_key_reports_yes_without_validation` pin to match, since that was the question it was waiting on. On your open point: for now an entered-but-unvalidated key reports `No` (it isn't usable yet and there's no fourth value in the spec), but that's a one-line change once PostNL confirms what they'd prefer.

**7 — `main` is merged in** and I dropped this branch's `bootstrap-unit.php` stub in favour of main's empty one, like you flagged.

Unit suite's green (694 tests). Two things I'd like your steer on:
- I left the SDK-path `NewKey` header out of scope for now — `get_new_key_header_value()` is correct on the legacy `Base` path, but the SDK builder doesn't expose a header hook, so wiring it onto the V4 requests feels like its own piece. Happy to fold it in if you'd rather keep it together.
- Joris's read-only field is still in this PR since it was already here — say the word and I'll split it into its own PR so this one stays purely V4 key-handling.

dustinparker:
Notes from the call with PostNL, plus the naming Joris signed off on this afternoon. All of it goes in this PR.

The screenshots are a static mockup, so treat them as the visual target rather than exact markup.

Settings screen with Old API Key read-only, API Key below it, and the four status states

## 0. Before any of the UI work: the validator has to come off the SDK

This one is on me. In my review on the 21st I said the validation call should go through the SDK to a V4 endpoint, because a V1 call cannot prove the key has V4 access. You implemented exactly what I asked for, and I was wrong. It breaks the release two ways.

**It cannot run in a release build.** The SDK is in `require-dev`, `.github/workflows/deploy.yml:146` builds with `--no-dev`, and `deploy.yml:236` fails the build outright if the SDK ends up in the zip. So in a merchant build `new BarcodeRequest(...)` raises `Error: Class not found`. The `\Throwable` block catches it, both `instanceof` checks return false because PHP does not autoload for `instanceof` against a missing class, and it returns `postnl_key_error`. The merchant sees "The newly entered API key is invalid" for a perfectly good key. `Exception_Converter` imports SDK classes too, so it degrades the same way.

That inverts the one thing this release exists to produce. The validated hash never gets set, the plugin never switches over, and `NewKey` reports `No` for exactly the merchants who did adopt.

**It is a parse error on PHP 7.4.** The `BarcodeRequest` call uses named arguments (`customerNumber:`, `customerCode:`), which are 8.0+, and `composer.json` pins `"php": ">=7.4"`. A 7.4 merchant saving settings with a key entered gets a fatal.

**The fix is to go back to a V1 Barcode call.** I was reasoning about a world where V4 is already live, but this ships on 14 September, before V4 is switched on, so the validator has to survive a build with no SDK in it. The validator is really answering two questions that have different deadlines:

| Question | When it is needed | How to answer it |
|---|---|---|
| Is this key live and functional? | 14 Sep, every merchant | V1 Barcode call. No SDK, 7.4-safe |
| Does this key have V4 entitlement? | Whenever PostNL opens V4 | `Service_Factory::has_v4_key()` plus the V4 test build |

The ClickUp spec only asks for the first: "A test call with they new key (for example a call to our Barcode API) should be sent out upon saving the configuration." `Legacy\Barcode\Client extends Base`, `Base::set_api_key()` is public, and `send_request()` already throws on both transport and API errors, so the V1 route is available without new plumbing.

Worth confirming every changed file passes `php7.4 -l` before this merges. Right now `Key_Validator.php` is the only one that fails.

## 1. Field renames

Joris agreed to drop the "new" framing. On a fresh install the only usable field was called "New API Key", which means nothing to someone who never had an old one.

| Option key | Was | Now (EN) | Now (NL) |
|---|---|---|---|
| `api_keys` | Production API Key | **Old API Key** | **Oude API Key** |
| `api_keys_new` | New API Key | **API Key** | **API Key** |
| `api_keys_sandbox` | Sandbox API Key | **Old Sandbox API Key** | **Oude Sandbox API Key** |
| `api_keys_sandbox_new` | New Sandbox API Key | **Sandbox API Key** | **Sandbox API Key** |

Joris asked specifically for "Oude" so BE merchants running the interface in Dutch get it translated rather than seeing the English word.

Labels only. The option keys do not move, so there is no migration and nothing to unpick if it changes again.

One trap worth knowing about: "Sandbox API Key" is now the label for a *different field* than it was before the update. A merchant who had a value under that label will find it under "Old Sandbox API Key" after updating. Only one environment renders at a time so they never appear together, but it is worth being deliberate about.

The tooltip sentences still need Dutch from Joris. The English in the mockup is mine and is a starting point, not final.

## 2. Old key field: hide when empty, read-only when set

`api_keys` is unconditionally read-only right now. That needs to become conditional on the saved value:

- Empty → do not render the field at all
- Non-empty → render read-only, the way it is today

New install with the old key field not rendered at all

Decide it on the **saved** value, not the posted one, or the field flickers on a failed save. And `api_keys_sandbox` has no readonly attribute at all today, so it needs the same treatment.

Read-only inputs still POST, so the stored key round-trips normally. The hidden case skips the save, and there is nothing to lose there by definition.

## 3. Status section

New row directly under the API Key field. I modelled it on the Checkout Rates status section in ShipStation, which solves the same problem: a bold coloured label, an em dash, one plain sentence, then a `description` paragraph saying what to do next.

Joris also confirmed the fourth `NewKey` value, so the states map one to one onto the header:

| Status | Colour | `NewKey` |
|---|---|---|
| Valid | `#008a20` green | `Yes` |
| Not valid | `#d63638` red | `Entered` |
| Same as old key | `#dba617` amber | `Same` |
| Not set | `#757575` grey | `No` |

**A key that fails validation still gets saved.** Joris asked about this directly, since he had assumed a failed validation would block the save and leave `No` as the only possible value. It does not, and that is deliberate: we store the value and show what is wrong underneath it, which is the whole point of the status section. So please do not add a guard that rejects the save on a validation failure. The key persists, the validated flag does not get set, and the header reports `Entered`.

**Derive the status and the header from the same method.** `get_new_key_header_value()` already computes this distinction, so the status section should read from it rather than running its own checks. If the two drift you get a settings screen saying "Valid" while every API call reports `No`, and that is very hard to debug from PostNL's side.

One consequence of the V1 fix in section 0, which is mine to settle with Joris rather than yours: with V1 validation, "Valid" and `NewKey: Yes` assert that a distinct key was entered and works on V1, not that it carries V4 entitlement. `Same` catches the most common version of that mistake and PostNL issues these keys specifically for V4, so the gap is narrow, but they are going to flip the migration based on that number so they should know what it proves. Build to the table above and I will confirm the wording with him.

Two details in the screenshots that are easy to miss:

"Not valid" covers more than a rejected key. Validation also fails when Customer Code and Number are not filled in, and when PostNL cannot be reached. In those cases the key may be perfectly fine, so the label can stay the same but the sentence underneath needs to differ. Telling someone their key is invalid when we simply could not check it is a support ticket.

"Same as old key" is unreachable on a new install, since there is no old key to match. Worth a test either way.

## 4. Two banners

`Api_Key_Banner::get_message()` branches on whether the old key field has a value. Everything else about the banner stays as it is, including dismiss and remind-later.

The two banner variants, one for existing merchants and one for new installs

Rick is writing both. The placeholder text in the screenshots is mine, so do not translate it.

## Still outstanding

The "get a key" link points at `mijn.postnl.nl` in the mockup. PostNL needs to confirm the real Self Service URL before this ships, since SSAM goes live on 14 September.

dustinparker:
This is looking good. I tested it on a live wp-env install in the release shape, meaning `composer install --no-dev` with no SDK in `vendor/`.

Six changes to request.

### 1. The screen after saving shows the state from before the save

Enter a valid key, save, and the page that comes back tells you that you have not entered one.

```
WITHOUT the fix -> status row: Not set | banner: shown  | NewKey: No  | key sent: OLDKEY999
WITH the fix    -> status row: Valid   | banner: hidden | NewKey: Yes | key sent: VALIDKEY-AAA
```

`PostNL::__construct()` reaches `Settings::get_instance()` via `init_form_fields()`, so the singleton has cached its settings before `process_admin_options()` runs. The parent call updates the DB and the method object but not the singleton, and the status row, the banner, `get_new_key_header_value()` and `get_effective_api_key()` all read through it for the rest of the request. It corrects itself on the next page load, which is why it is easy to miss locally.

`Settings::get_instance()->init_settings();` after `parent::process_admin_options()` fixes all four. That is the second line above.

### 2. The status should check the key on blur

The fix above only makes it right after saving. It is still stale while the merchant is typing, which is when they want to know.

- New AJAX endpoint, `manage_woocommerce` then nonce, calling `Key_Validator::validate()`. Read-only, writes no option and no hash, and does not log the key.
- Take the typed Customer Code and Number from the request, not from storage, or someone filling in all three fields at once gets told to fill in details they just entered.
- Split `get_new_key_status()` into a `build_new_key_status( $new_key, $original, $validated, $is_sandbox, $saved )` that only maps state to copy. The render path and the endpoint both call it, so the screen and the header still cannot disagree.
- A checked but unsaved key needs its own amber state, "Works, not saved yet — This key works. Save changes to start using it." Green "Valid" stays for saved and in use, otherwise the row says Valid while `NewKey` still reports `Entered`.
- Empty and same-as-old-key are decidable in the browser, so only a distinct key should call out.
- Every successful Barcode call mints a barcode from the merchant's range, so memoise checked values per page load, abort in-flight requests, and rate limit server side. I will tell PostNL the volume goes up rather than switch endpoints.

### 3. A PostNL outage tells the merchant their key is invalid

`Key_Validator::validate()` inspects the body before the status code, so any non-2xx carrying an Apigee `fault` reads as a key problem:

```
429 + fault (quota exceeded)       says: The newly entered API key is invalid.
503 + fault (backend unavailable)  says: The newly entered API key is invalid.
200 + no Barcode field             says: The newly entered API key is invalid.
```

The other half is that `postnl_key_http_status` catches everything else non-2xx (`Key_Validator.php:122-128`) and `PostNL.php:125` treats the lot as transient, so a 400 and a 404 both tell the merchant to retry something that cannot succeed.

Classify on the status code first and use the body only for the message. Four reasons, each with its own sentence:

| Condition | Reason | Sentence |
|---|---|---|
| 401, 403, or a fault/`Errors` body on a 2xx | `invalid` | PostNL rejected this key |
| Customer Code or Number empty | `missing` | We could not check this key because those are not filled in |
| transport failure, 5xx, 429, or a 2xx with no Barcode | `unreachable` | We could not reach PostNL, try saving again in a few minutes |
| any other 4xx | `rejected` | PostNL could not process the check, most likely Customer Code or Number does not match this key |

Right now `unreachable` and `missing` produce identical copy, so a real outage leads with "Fill in your Customer Code and Customer Number" to someone who already has. Please keep the mapping in one method that both the save path and the new endpoint call, so the two cannot drift.

One comment to fix while you are there. `PostNL.php:122-124` says the transient branch stops us disabling a working key on a blip. It does not: the early return at `PostNL.php:110` makes zero HTTP calls for an unchanged validated key, so nothing reaches that branch in the case described. The branch is still worth keeping, but the reason is that it avoids a false red. The comment also says "5xx" when the code catches far more.

### 4. The status copy assumes an old key exists

Wording only, the logic underneath is correct in every combination I tested. On a fresh install `get_effective_api_key()` correctly returns `''`, and the row still says:

> PostNL rejected this key, so the plug-in is still using your old key.

Nothing is in use there, so no labels and no checkout options, which is worse than the sentence suggests. "Not set" has the same problem and no branch at all: it says the key is needed "before the new PostNL APIs are switched on" when a new install needs it immediately.

Three strings, `Settings.php:1210`, `1216` and `1184`. `get_new_key_status()` already has the right value for the requested environment at line 1144, so pass it down rather than calling `get_original_api_key()` inside the copy helper, which reads the current environment while the row renders for both. No banner copy changes, that is Rick's.

### 5. Nothing added this round has a test

`SettingsNewApiKeyTest` is the only file referencing any of this code, and everything it covers is from the previous two rounds. `Key_Validator`, `process_new_api_key_validation()`, `get_new_key_status()`, the field gating and the banner have no tests at all. Every defect above is in untested code.

Unit: the status mapping across all four header values and the four reasons, with and without an old key; the error code to reason classifier; `Key_Validator::validate()` across the response matrix in item 3; and "Same as old key" being unreachable with no old key, which is the one I asked for last time.

Integration: the post-save render from item 1, and the field gating. Item 1 has to be an integration test, because a unit test on `get_new_key_status()` passes while the screen is still wrong. `ServiceFactoryRecursionTest` is the precedent.

### 6. Changelog and PR description

`readme.txt`, `changelog.txt` and `README.md` are untouched and have no pending entry. The convention is for the PR to add it with a placeholder date, the way #342 did.

The description is also out of date, and the changelog line is written from it. It still says validation goes through the SDK against V4, and it still uses the old field names, the old read-only behaviour and three `NewKey` values. As it stands the release notes would describe fields that do not exist.

I would put this under 5.9.11. #345 renamed the last pending 6.0.0 entry down to 5.9.10, which reads like 6.0.0 is being kept for V4.

On your question about the SDK path, leave `NewKey` off it. The legacy calls are where the signal is, since a V4 request can only happen once the key has already validated.

dustinparker:
Two things to fold into this PR before we release, ideally today or tomorrow.

**Changelog version.** 5.9.11 went out this morning with the WordPress 7.1 and WooCommerce 11.1 compatibility tweak, so the entry here needs to move to 5.9.12 in `changelog.txt`, `readme.txt` and `README.md`. Main also needs merging in, and it will conflict on those three files.

**wp.org packaging.** Trunk is publishing files we do not ship in the zip: `AGENTS.md`, `CLAUDE.md`, `docs/`, `phpunit-unit.xml.dist`, `phpunit-integration.xml.dist`, `.wp-env.json`, plus a stale `vendor/myparcelnl`. They are all in `.distignore` already and the release zip is clean, so the cause is on the SVN side. Rsync does not delete files in the destination that are excluded, so anything that landed in trunk before the `.distignore` entry stays there, and every `svn copy` tag carries it forward. The agent docs have been in every tag since 5.9.6 and `vendor/myparcelnl` since 5.9.5.

Would you be able to add `--delete-excluded` to the trunk rsync in `deploy.yml`, `svn delete` the stale paths from trunk, and extend the existing ZIP verification step to assert against the SVN working copy before the commit as well? `vendor/myparcelnl` is the odd one, since it is not in `.distignore` and was never a dependency here, so it is worth checking how it got in.

PR #312, opened 2026-05-10
  Fix house number lost in blocks checkout when address validation is…

sal4sup:
… disabled

### All Submissions:

Closes # .
https://app.clickup.com/t/868jgz4zv

1.Disable "Validate Dutch addresses " and "Use PostNL address-field "
2.Go to the checkout and add 1 line address contains house number example "Diergaardesingel 67" 
3.try to fill address with autofill 

### Other information:

### Changelog entry


Abdalsalaam:
## 🤖 Automated PR review
**ClickUp task:** [House number disappearing with address fields disabled](https://app.clickup.com/t/868jgz4zv)

This is a correct, minimal, well-scoped fix. The root cause is confirmed: `Address_Utils::set_post_data_address()` → `split_address()` strips the house number out of `shipping_address_1` (e.g. `Diergaardesingel 67` → `Diergaardesingel`, see `src/Address_Utils.php:122,130`). The old code then wrote that stripped value to `WC()->customer->set_shipping_address_1()`, so on re-render the single-line field lost its number. Capturing the pre-split value and restoring it only on the disabled-fields path fixes exactly this.

**Spec coverage**
- ✅ House number persists in **blocks** checkout — fix lives in the blocks-only AJAX handler `handle_set_checkout_post_data` (`Extend_Block_Core.php:369`); blocks JS posts `shipping_address_1` (`block.js:442`).
- ✅ Targets the **disabled** "Use PostNL Fields" path specifically — restore is gated by the `else` of `is_reorder_nl_address_enabled()` (`Extend_Block_Core.php:486-490`).
- ✅ No regression to the **enabled** path — when reorder is enabled, both the validated and else branches still use the split street-only value (`:477, :487`).
- ✅ No session fragmentation reintroduced — change only adjusts the value passed to `set_shipping_address_1`; no new session keys.
- ⚠️ Browser-autofill / whole-form-clearing trigger — not handled explicitly, but the autofilled value flows through the same handler, so the server-side root-cause fix covers it. Worth confirming in manual testing.

**Findings**
- 🟢 **[low] robustness/implicit-contract** — `$original_shipping_address_1` is captured at `Extend_Block_Core.php:385` *before* `set_post_data_address()`. That function's billing-copy branch (`Address_Utils.php:46-56`) only runs when `ship_to_different_address` is empty; the blocks JS hardcodes `ship_to_different_address: '1'` (`block.js:448`), so the capture is safe today. But the disabled path now silently depends on that JS invariant — a future caller posting without `ship_to_different_address='1'` could restore an empty/stale `address_1`. — *fix:* fall back to the post-split value when empty, e.g. `set_shipping_address_1( '' !== $original_shipping_address_1 ? $original_shipping_address_1 : ( $sanitized_data['shipping_address_1'] ?? '' ) )`.
- 🟢 **[low] testing/no-coverage** — no automated test exercises `handle_set_checkout_post_data` or the disabled-fields restore; this regression class can recur silently (`tests/`). — *fix:* add manual test notes for all four matrix cells (PostNL-fields on/off × validation on/off); optionally a PHP unit test asserting the inline house number survives with reorder disabled.
- ℹ️ **[info] docs/wording** — the changelog says the bug occurs only when *both* "Use PostNL address-field" **and** "Validate Dutch addresses" are disabled. The `else` branch (`:486`) is also reached when validation is enabled but returns nothing/fails while PostNL fields are off, so the fix actually covers slightly more. Optional reword to focus on the disabled PostNL-fields root cause. No code change.

**Testing notes**
1. **Blocks checkout, PostNL fields OFF + validation OFF** (the reported case): enter a NL address with inline house number (e.g. `Diergaardesingel 67`), trigger a delivery-options refresh, and confirm the house number is still present in `shipping_address_1` after re-render.
2. **Blocks checkout, browser autofill**: let the browser autofill the address line, then confirm the number survives the subsequent form refresh.
3. **Blocks checkout, PostNL fields ON** (enabled path regression check): confirm street and house number still render in their separate fields and the single-line `address_1` shows street only.
4. **Blocks checkout, validation ON + PostNL fields OFF**: confirm the validated `street + house_number` is still combined correctly (`:480-482`) and nothing double-appends.
5. Classic checkout is on a separate code path (`Frontend/Container.php`) and untouched — a quick smoke test confirms no collateral change.

PR #319, opened 2026-06-01
Fix Belgium country mapping in Invullen met PostNL autofill

sal4sup:
### All Submissions:

Closes # .
https://app.clickup.com/t/4550855/868jqbryp

  1. Enable "Invullen met PostNL" and connect a Belgium PostNL account.
  2. Go to checkout and click the "Invullen met PostNL" button.
  3. Confirm the country field is set to Belgium (BE) and no postcode validation error appears.

### Other information:

### Changelog entry

> Fix incorrect country (NL) being set when autofilling checkout with a Belgium PostNL account.


Abdalsalaam:
## 🤖 Automated PR review
**ClickUp task:** [Invullen met PostNL (BE account): Country is wrong](https://app.clickup.com/t/868jqbryp)

The approach is sound and correctly scoped. The original bug was the hardcoded `?? 'NL'` fallback firing when PostNL returns a localized country name (e.g. "België"/"Belgique") that WC's English-only country list can't match. The new `resolve_country_code()` prefers the machine-readable `countryCode`, then a multilingual NL/BE name table, then the WC name lookup. A single PHP change covers both checkout modes since both JS paths read the server-resolved `primaryAddress.countryName`. The only real weakness is test coverage.

**Spec coverage**
- ✅ **BE account sets country = BE** — `resolve_country_code()` maps `countryCode` `BE` and localized names (`belgie`/`belgië`/`belgium`/`belgique`/`belgien`) to `BE`.
- ✅ **NL accounts still map to NL** — `countryCode` `NL` and `nederland`/`netherlands` map to `NL`; English WC name covered by `get_country_code_by_name`.
- ✅ **Works in classic and blocks checkout** — single PHP change writes the resolved code into session `countryName`; both `assets/js/fill_in_with_postnl.js` (L66,73) and `client/checkout/postnl-fill-in-with/block.js` (L67) consume it.
- ❓ **No postcode-validation error for BE** — logically follows from `country=BE`, but no automated test/evidence added; relies on the manual test steps.
- ⚠️ **Mapping is robust (not a hardcoded NL fallback)** — greatly improved, but a silent unconditional `return 'NL'` remains and the name table is a finite allowlist with no `trim()`/accent normalization.
- ⚠️ **Test coverage for the fix** — added JS test bypasses PHP; `resolve_country_code()` itself is untested.
- ✅ **Changelog updated** — entries added to `README.md`, `changelog.txt`, `readme.txt`.

**Findings**
- 🟠 **[High] test-does-not-exercise-fix** — The new test "should set country to BE" (`tests/js/postnl-fill-in-with.test.js` ~L427-470) mocks `fetch` so `data.primaryAddress.countryName` is already the literal `'BE'`; `block.js:67` reads it straight through. Jest never runs PHP, so `resolve_country_code()` — the entire fix — has zero automated coverage (repo has no PHPUnit suite). The test would still pass if someone re-added an early `NL` fallback. The fixture is also unrealistic: feed `countryName:'BE'` through the real PHP path and it returns `'NL'` (no `countryCode`, `'be'` not in the name table, no WC English-name match) — so the fixture doesn't match a payload that would actually have triggered the original bug. — fix: drive the real server path with a localized name (e.g. `countryName:'België'`/`'Belgique'`) and/or add a thin PHP test asserting `resolve_country_code(['countryName'=>'België'])==='BE'`, `['countryName'=>'Nederland']==='NL'`, `['countryCode'=>'BE']==='BE'`; at minimum note in the PR that the PHP mapping was only manually tested.
- 🟡 **[Medium] hardcoded-NL-fallback** — `resolve_country_code()` ends with an unconditional `return 'NL';` (`src/Frontend/Fill_In_With_Postnl_Handler.php`). The reviewer notes explicitly require the fix to not be a hardcoded NL fallback. A BE address whose `countryName` is an unanticipated spelling/locale not in the table AND has no `countryCode` is silently coerced to NL — reproducing the exact reported bug (and the postcode error) for that variant. — fix: confirm from a real `get_postnl_user_info` payload whether `countryCode` is reliably present (the locale-independent primary path); if so the NL default is acceptable but should at least `Logger`-log unmapped names so locale gaps are observable; if not, return `''` and let WC surface a normal validation error.
- 🟢 **[Low] unvalidated countryCode passthrough** — `return strtoupper( $address['countryCode'] )` returns the code without validating it against `WC()->countries->get_countries()`, so a malformed value is stored verbatim as the WC billing/shipping country. — fix: validate the upper-cased code against `array_keys( WC()->countries->get_countries() )` before returning; otherwise fall through to name resolution.
- 🟢 **[Low] missing name normalization** — name lookup uses `mb_strtolower($name)` with no `trim()`/accent normalization, so ` Belgium `, NBSP, or a differently-normalized `ë` (composed vs decomposed) miss the table and fall through to NL. — fix: `trim()` (and ideally a remove-accents pass) before the lookup. Minor since `countryCode` is the primary path.
- 🟢 **[Low] WPCS docblock missing summary** — the docblock above `resolve_country_code()` jumps straight to `@param` with no one-line summary, unlike sibling methods (e.g. `get_country_code_by_name`). — fix: add a summary line such as `Resolve a two-letter WooCommerce country code from a PostNL primaryAddress.`

**Testing notes**
- Blocks checkout: log in via Fill In With PostNL using a **BE** PostNL account and confirm the country select shows Belgium with no NL postcode-validation error.
- Classic checkout: repeat with a BE account; verify `#billing_country`/`#shipping_country` resolve to BE and trigger the address listeners correctly.
- Regression: repeat with an **NL** account in both modes and confirm country stays NL.
- Edge case: simulate a `get_postnl_user_info` payload whose `countryName` is a localized BE spelling with **no** `countryCode` (mirroring the original bug) and confirm it still resolves to BE — this is the scenario the added JS test does not cover.
- Confirm with PostNL API docs whether `user_info` returns `countryName` as a localized name or ISO code, and whether `countryCode` is reliably present, then align the fixtures and the fallback accordingly.

PR #321, opened 2026-06-02
Extract V1 product-code mapper into pure class

sal4sup:
### All Submissions:

  Move the product-code mapping data out of `Helper/Mapping.php` into a new
  pure PHP class `Helper/Product_Mapper/V1_Mapper.php`. `Mapping.php` keeps
  all its existing public methods but now delegates to `V1_Mapper` instead of
  holding the data itself. No shipping behaviour, product codes, or plugin
  output changed.
  
Closes # .
https://app.clickup.com/t/4550855/868jqaqew

1.
2.
3.

### Other information:

### Changelog entry


Abdalsalaam:
## 🤖 Automated PR review
**ClickUp task:** [4. Extract V1 product-code mapper + exhaustive unit tests](https://app.clickup.com/t/868jqaqew)

Clean, well-scoped extraction. I independently verified the core acceptance criterion by running both versions side by side: all six data methods produce **byte-identical output** between `origin/main`'s `Mapping` and the new `V1_Mapper` (`products_data`, `EU_ROW_products`, `european_shipment_products`, `globalpack_products`, `shipping_return_labels_options`, `additional_product_options` — all `===`). `V1_Mapper` is genuinely pure (zero WP calls in the diff), and `Mapping.php` is a thin delegate. `products_data()` is exhaustively golden-mastered: 88 live entries == 88 provider cases, 0 ambiguous sorted-combination collisions, so `find_entry()`'s order-independent matching is safe. The one real gap is that the two ancillary code+characteristic-bearing methods are spot-checked, not pinned.

**Spec coverage**
- ✅ **Pure `V1_Mapper` class, no WP calls/side effects** — verified: no `apply_filters`/`get_option`/`__()`/`wp_*`/`sanitize`/`$wpdb`/`new`/`global` in the added file; static data only.
- ✅ **`Mapping.php` delegates with zero behavior change** — all 6 data methods now `return V1_Mapper::x()`; runtime output proven identical to `main` for every method.
- ⚠️ **Unit test enumerates every zone × option combo asserting code + characteristics (golden-master)** — `products_data()` is fully pinned (88/88, hardcoded literals, not derived from `V1_Mapper`). But `shipping_return_labels_options()` (carries product codes + characteristics) is only count-checked, so not *every* code+characteristic mapping is pinned.
- ⚠️ **Acceptance: 100% of combinations covered by assertions** — 100% for `products_data()`; the `shipping_return` 15-code lists and their characteristic options are not value-asserted.
- ✅ **Acceptance: `Mapping.php` output unchanged** — independently confirmed byte-faithful (runtime `===` for all 6 methods).

**Findings**
- 🟡 **[Medium] test-coverage gap** — `shipping_return_labels_options()` is not golden-mastered: `NL/NL shipping_return['products']` (15 codes) is only `assertCount(15)`, its `['options']` (char `152`/opt `026`) is never asserted, and `return_all_labels_not_active['products']` (15 codes) + `['options']` (`152/026` and `191/004`) are never value-checked (only the option *count* is). A reorder or single-code typo in these lists passes the suite, and the spec explicitly names "characteristics." `tests/php/unit/Helper/Product_Mapper/V1_MapperTest.php:2041-2052` — fix: add `assertSame` on the full `shipping_return['products']` + `['options']` and `return_all_labels_not_active['products']` + `['options']` arrays (verbatim codes), matching the pinning `products_data()` already gets.
- 🟢 **[Low] documentation/parity** — the inline comment `// Empty mean it can be combined with all products.` on the `NL/NL in_box 'products' => array()` line was dropped in the extraction (the only non-whitespace difference across all six data blocks; zero behavior impact, but it documented a non-obvious semantic). `src/Helper/Product_Mapper/V1_Mapper.php:1304` (in_box `'products' => array()`) — fix: restore the comment for parity.

**Testing notes**
- Run `npm run test:php:unit` (phpunit-unit.xml.dist) and confirm green before merge — dev deps (`brain/monkey`, `phpunit`) are Private-Packagist-gated, so the suite could only be verified statically/by direct execution of the data methods locally; confirm the CI unit job (composer install with `COMPOSER_AUTH`, then phpunit) passes.
- Sanity-grep the new class for purity if touched again: it must call no WP functions (it currently calls none).
- After adding the suggested `assertSame` assertions, intentionally reorder one code in `shipping_return['products']` and confirm the test now fails (it currently would not).
- This is data-only/admin label-mapping; no checkout UI change, so classic-vs-blocks checkout parity is not affected by this PR.
- Optional: add a one-line PR note that the inactive commented-out `NL/NL delivery_day` `insured_shipping → 3087` block was intentionally not carried over (it was never an active entry — no behavior change).

dustinparker (approved):
Everything is working as expected. The mapping data moves into the pure `V1_Mapper` with `Mapping` delegating to it, and every product code, option, and array order comes out identical to before, so nothing downstream changes. The golden-master tests pin all 89 combinations, which gives the V4 mapper a concrete contract to match.

PR #322, opened 2026-06-02
Postnl v4 v4 product mapper

sal4sup:
### All Submissions:

Adds a pure PHP mapper class that translates a V1 product combination (origin, destination, flow, options) into a V4 ShipmentType + Services shape, or a legacy-only marker when no V4 equivalent exists yet.
  
Closes # .
https://app.clickup.com/t/4550855/868jqaqey

1.
2.
3.

### Other information:

### Changelog entry


Abdalsalaam:
## 🤖 Automated PR review
**ClickUp task:** [5. Build V4 mapper (ShipmentType + Services) + tests](https://app.clickup.com/t/868jqaqey)
**Verdict:** 🟡 Needs changes

Reviewed only the incremental V4-mapper diff (`git diff origin/postnl-v4-v1-product-mapper origin/postnl-v4-v4-product-mapper`): `src/Helper/Product_Mapper/V4_Mapper.php` (+406) and its test (+930). The domestic mapping work is solid — O(1) indexed 88-row matrix, order-independent option keying, exhaustive data-provider with runtime count assertions. The international half of the spec and one explicit Open Item are unmet.

**Spec coverage**
- ✅ `has_v4_equivalent($combination): bool` present (`V4_Mapper.php:61`), delegates to `map()`.
- ✅ All 10 "Not yet available" codes flagged no-V4-equivalent — 7 with matrix entries marked `$nya` (`V4_Mapper.php:267,291,329-333`); 1175/3574/4983 have no V1 row and are held in `NOT_YET_AVAILABLE_CODES` (`:37`). Runtime test asserts `has_v4_equivalent()===false` for the set.
- ✅ ShipmentType + domestic Services flags emitted (`statedAddressOnly`, `returnWhenNotHome`, `deliveryConfirmation`, `insuredValue`).
- ❌ International `services.bundle` value — **no `bundle` key anywhere** in the diff.
- ⚠️ `internationalShipmentData` hints — key exists but is always the empty-array default; every EU/ROW row is `legacy_result(... $nc)`, so the international V4 shape is never produced or tested.
- ❌ Open Item: FLAG the SDK Services-DTO `bundle` gap — no comment / TODO / marker referencing it exists. Gap was silently dropped.
- ⚠️ "Assert a shape for EVERY task-4 combination, no silent gaps" — provider is hand-transcribed, not derived from `V1_Mapper::products_data()` (which is present on-branch).

**Findings**
- 🟠 **[High] spec-gap** — Open Item required FLAGGING the missing SDK `services.bundle` property; instead it is silently absent (no `bundle`, no TODO/FIXME). All international rows route to Legacy so the gap is invisible in code `V4_Mapper.php` (class PHPDoc `:33`, `v4_result()` `:129-146`, intl branches `:294-324,381-400`) — fix: add a discoverable marker (class-PHPDoc note + `// TODO(SDK): Services DTO has no bundle property; international blocked on SDK update — keep on Legacy`), ideally a test asserting the marker so it can't be removed silently.
- 🟡 **[Medium] spec-coverage** — international deliverable (`internationalShipmentData` hints + `bundle`) is structurally present but never populated; `internationalShipmentData` is always `array()` `V4_Mapper.php:144` (no `v4_result` row in `:294-324,381-400` passes a 6th arg) — fix: either populate at least one international row or document the deferral explicitly, tied to the SDK-gap flag, so the empty key reads as a known deferral not an oversight.
- 🟡 **[Medium] dead-code vs stated intent** — `NOT_YET_AVAILABLE_CODES` PHPDoc claims it "prevents accidental mapping," but `map()` never reads the constant; protection for the 3 entry-less codes comes solely from matrix absence `V4_Mapper.php:37-48` vs `map()` `:71-101` — fix: make it load-bearing (after resolving `$row`, force `has_v4_equivalent=false`/`REASON_NOT_YET_AVAILABLE` if the code is in the set) so no future matrix edit can leak those codes to V4.
- 🟡 **[Medium] test-coverage** — the 88-row provider is hand-copied, not enumerated from `V1_Mapper::products_data()` (present on-branch), and the cited `docs/.../v4-mapper-combination-matrix.csv` (`V4_Mapper.php:33`, test `:29,374`) exists on no branch (v1/v4/main) — a combination dropped from the provider would pass CI undetected. Fix: add a test that enumerates `V1_Mapper::products_data()` and asserts `map()` returns a known result for each; commit the CSV or drop the citations.
- 🟢 **[Low] partial spec coverage** — 3 of the 7 spec'd Services flags (`registered`, `minimalAgeCheck`, `deliveryWindow`) are never emitted; `minimalAgeCheck` (id_check/18+) rows all go to Legacy `V4_Mapper.php:255-262,267-268,354` — fix: if intentionally deferred, document it so reviewers can tell "deferred" from "forgotten."
- 🟢 **[Low] robustness** — `options_key()` sorts case-sensitively with no trim/lowercase; mixed-case/whitespace input falls through to `REASON_UNKNOWN_COMBINATION` `V4_Mapper.php:110-116` — fix: document the canonical-lowercase precondition, or normalize before `sort()` and add a test.

**Testing notes**
- Run the new suite: `composer test -- --filter V4_MapperTest` (or the project's PHPUnit harness) and confirm the 28-true / 60-false runtime counts hold.
- Add a runtime assertion that iterates `V1_Mapper::products_data()` and checks every combination resolves through `V4_Mapper::map()` — this is the real "no silent gaps" guard the spec asks for.
- Add a negative test: feed an options array in non-canonical case/whitespace and assert the resolved row (will currently fail → confirms the `options_key` gap).
- Confirm the chosen SDK-gap marker (comment/constant) is covered by an assertion so it survives refactors.
- No checkout/runtime wiring in this diff (pure mapper), so classic-vs-blocks checkout testing is N/A here; verify when Service_Factory consumes this in a later stacked PR.

dustinparker (approved):
Everything is working as expected. Every combination from the V1 mapper resolves to either a confirmed V4 shape or an explicit Legacy-only reason with no silent gaps, the not-yet-available codes stay on Legacy, and the international rows are correctly deferred since the SDK's `Services` DTO really has no `bundle` property.

One note to carry into task 8: the mapper emits `<order_total>` and `<from_selected_location>` as substitution tokens, and the two behave differently if a caller forgets to replace them. `insuredValue` is typed `?float` in the SDK, so an unsubstituted token throws a TypeError immediately. `pickupLocationId` is typed `?string`, so the literal `<from_selected_location>` passes straight through and would end up in the label request. That applies to the six V4 pickup rows (21, 22, 66, 67, 75, 76). It would be worth documenting the convention in the class docblock, or exposing the token prefix as a constant so the consumer can assert that nothing `<`-prefixed survives substitution.

PR #323, opened 2026-06-05
Add 7 Rest_API\Contracts service interfaces for Legacy/V4 transport a…

sal4sup:
…bstraction

### All Submissions:

  Adds 7 PHP interface files under `src/Rest_API/Contracts/` — the compile-time
  service contracts that both the existing Legacy transport and the upcoming V4
 transport must satisfy.  No implementations, no wiring, no behavior changes.

  Interfaces added:
- `Barcode_Service_Interface` — `generate(array $post_data): array`
- `Label_Service_Interface` — `create(array $post_data): array` (returns normalized `_postnl_order_metadata['labels']` shape)
- `Return_Label_Service_Interface` — `create(array $post_data): array` + `activate(int $order_id): array`
- `Timeframe_Service_Interface` — `get_delivery_options(array $post_data): array`
- `Pickup_Location_Service_Interface` — `get_pickup_locations(array $post_data): array`
- `Postcode_Check_Service_Interface` — `check(array $post_data): array` (Legacy-only permanently; included for factory uniformity)
- `Smart_Returns_Service_Interface` — `generate(\WC_Order $order): array`

All signatures and return shapes are derived directly from current callers (`Order\Base`, `Order\Single`, `Frontend\Container`) — no signatures were invented.

Closes # .
https://app.clickup.com/t/4550855/868jqaqf2

1.
2.
3.

### Other information:

### Changelog entry


Abdalsalaam:
## 🤖 Automated PR review
**ClickUp task:** [6. Define the 7 service interfaces](https://app.clickup.com/t/868jqaqf2)

Clean, correctly scoped contracts-only PR. `gh pr diff 323` and the merge-base diff against `main` both show exactly the 7 interface files (631 additions, 0 deletions, 1 commit) — no implementations, no wiring, no parent-stack noise. All 7 files lint clean (`php -l`), are pure `interface` declarations, and carry the ABSPATH guard. Signatures and return shapes were verified line-by-line against the real callers. The findings below are documentation-precision nits only; none block merge.

**Spec coverage**
- ✅ New `src/Rest_API/Contracts/` directory with exactly 7 interfaces, names matching the spec (Barcode, Label, Timeframe, Pickup_Location, Return_Label, Postcode_Check, Smart_Returns).
- ✅ Signatures derive from real current usage — verified against `Order\Base::create_barcode`/`maybe_create_return_label`, `Order\Single::postnl_send_smart_return_email` (`ResponseShipments`/`PrintcodeLabel`) and `postnl_activate_return_function` (`successFulBarcodes`), `Frontend\Container::get_checkout_data` (`DeliveryOptions`/`PickupOptions`) and `validated_address` (`$response[0]['city'/'streetName'/'houseNumber']`). No invented signatures.
- ✅ PHPDoc documents the return shape for each method; endpoint references match the real `$endpoint` constants.
- ✅ Postcode_Check interface present with explicit note that it stays V1-only permanently for factory uniformity.
- ✅ Acceptance met: interfaces compile, PHPDoc present, NO implementations leaked in. PSR-4 namespace (`PostNLWooCommerce\Rest_API\Contracts`) + file naming correct.

**Findings**
- 🟢 **Low / doc-accuracy** — `Pickup_Location_Service_Interface` PHPDoc says pickup count is "currently hardcoded to 20 via `Settings::get_number_pickup_points()`", but that method (`src/Shipping_Method/Settings.php:1102`) returns `3`. `src/Rest_API/Contracts/Pickup_Location_Service_Interface.php:~38` — fix: change `20` to `3`, or reword to "a fixed count (see `Settings::get_number_pickup_points()`)" to avoid coupling the doc to a literal.
- 🟢 **Low / doc-accuracy** — `Barcode_Service_Interface::generate()` documents `customer_code` as an optional caller-supplied `$post_data` key, but no caller passes it: `create_barcode()` (`Order/Base.php:741`) builds `$data` from `order`/`saved_data` only, and `maybe_create_return_barcode()` (`Order/Base.php:864`) derives it internally from `Settings::get_return_customer_code()`. `src/Rest_API/Contracts/Barcode_Service_Interface.php:~53` — fix: drop the `customer_code` `$post_data` key, or note it is an internal/derived value, not a caller input.
- ℹ️ **Info / doc-precision** — `merged_files` is documented as "Present only when label format is A4". `Order\Base::maybe_merge_labels()` (`Order/Base.php:902`) only returns the un-merged single-entry shape when there is exactly one label **and** format is A6; every other case (A4, or multi-collo A6) merges and adds `merged_files`. `Label_Service_Interface.php:~174` and the same wording in `Return_Label_Service_Interface.php:~452` — fix: reword to "Present whenever labels were merged, i.e. any case other than a single A6 label (covers A4 and multi-collo)."

**Testing notes**
- This is interface-only; verify implementer-facing correctness via the docs themselves: confirm the three doc nits above are reconciled against `Settings::get_number_pickup_points()` and `Order\Base::maybe_merge_labels()` before the Legacy/V4 implementations are written, so adapters don't return the wrong shape.
- When the Legacy adapters land, confirm `Barcode` returns the full decoded array (`['Barcode' => ...]`) while existing string-consuming callers (`create_barcode` returns the bare `$response['Barcode']`) are bridged — don't let an adapter return the bare string.
- When `Timeframe`/`Pickup_Location` adapters land, confirm they return the inner Checkout response body (the value stored under `['response']` by `get_checkout_data()`), not the `{response, post_data, letterbox}` wrapper.
- No checkout/runtime behavior changes in this PR; classic vs blocks checkout testing applies only once implementations + factory wiring land.

dustinparker (approved):
Everything is working as expected. The signatures are derived from the current callers rather than invented, including Smart Returns taking a `WC_Order` directly where the others take the post-data array, and each method documents the return shape the existing callers already consume. Nothing implements these yet, so there is no runtime impact.

One note to carry into task 7: the interfaces declare `: array`, but the legacy transport cannot guarantee that today. `Rest_API\Base::send_request()` returns `json_decode( $body_response, true )` with no array guard, so it is null on an empty or malformed body and a string on a JSON scalar. Callers currently absorb that, since `empty( $response['Barcode'] )` on null is true and the merchant ends up with the plugin's own "Cannot create the barcode" exception. If the legacy implementations pass `send_request()` straight through, that same response becomes an uncaught TypeError instead, which turns a handled error into a fatal. The `: array` contract is the right call, it just means the implementations need to normalize with something like `is_array( $response ) ? $response : array()` rather than pass the value through.

PR #324, opened 2026-06-06
Postnl v4 legacy clients interfaces

sal4sup:
### All Submissions:

This PR implements Subtask 7: Legacy clients implement the service interfaces from Subtask 6.

This PR is based on the previous service-interfaces branch / PR.

Changes included:

- Move the existing REST API Client.php and Item_Info.php files into the Legacy namespace.
- Keep backward compatibility for the old namespaces using class_alias shims.
- Add interface-compatible Legacy service wrappers.
- Keep existing callers unchanged.
- No Service_Factory wiring yet.
- No V4 SDK logic.
- No Router or Settings changes.
- No intended behavior change.

Extra implementation note:

Most legacy clients are raw API clients. They still need request-specific Item_Info objects, so they are not ideal factory-returned services by themselves.

For that reason, this PR adds stateless Legacy service wrappers such as Barcode_Service, Checkout_Service, Postcode_Check_Service, and Smart_Returns_Service. These wrappers build the existing Item_Info + Client internally and keep the current behavior unchanged.

Label, letterbox, and return-label flows also use service wrappers because their interfaces return the normalized label metadata shape stored in _postnl_order_metadata['labels'], not the raw PostNL ResponseShipments response.

Closes # .
https://app.clickup.com/t/868jqaqf8

1.
2.
3.

### Other information:

### Changelog entry


Abdalsalaam:
## 🤖 Automated PR review
**ClickUp task:** [7. Legacy clients implement interfaces](https://app.clickup.com/t/868jqaqf8)

Verified against the incremental diff vs `origin/postnl-v4-service-interfaces` (~39 files). The move is clean and mechanical: each `<Flow>/Client.php` and `Item_Info.php` body is byte-for-byte identical to the original (diff shows only `namespace`/`@package` lines change), original paths became pure `class_alias()` shims so no callers were touched, and subclass clients correctly repoint `use ...Shipping` → `use ...Legacy\Shipping`. All flagged issues are latent risks in the *new* code, not behavior regressions in this PR.

**Spec coverage**
- ✅ Move `Client`/`Item_Info` into `Legacy/<Flow>/` with updated namespace — all 8 flows moved; bodies byte-identical.
- ✅ ZERO logic change — confirmed via diff on the largest files (Shipping Client+Item_Info, Checkout Item_Info): only namespace/`@package` differ.
- ✅ `implements <Flow>_Service_Interface` on raw clients — Barcode, Checkout (Timeframe + Pickup), Postcode_Check, Smart_Returns implement directly; Letterbox/Return_Label/Shipment_and_Return are handled by the Label/Return wrappers per the PR note (intentional).
- ✅ Thin method builds `Item_Info` + calls `send_request()` returning existing shape — present on all raw-data clients (success-path shape preserved; see 🟡 below for the null-body edge).
- ✅ Leave callers untouched + `class_alias()` shims — diff touches only `src/Rest_API/`; 16 original paths are alias-only shims.
- ✅ Stateless service wrappers added (Barcode/Checkout/Postcode_Check/Smart_Returns/Label/Letterbox/Return_Label) — raw-data ones zero-arg constructable; Label/Letterbox/Return_Label extend `Order\Base` with no-op `init_hooks()` and delegate to `create_label`/`maybe_create_letterbox`/`maybe_create_return_label`.
- ✅ Wrappers return the normalized `_postnl_order_metadata['labels']` shape — delegated to the existing `Order\Base` pipeline, no duplication.
- ⚠️ Acceptance "all existing tests pass / type-checks against its interface" — behavior unchanged (new methods are not yet wired; cut-over is task 9), but **no test** asserts interface conformance or alias resolution, so CI does not exercise this criterion.

**Findings**
- 🟡 **Medium — behavior-drift / type-safety** The new thin methods/wrappers declare `: array` and `return $client->send_request();`, but `Base::send_request()` ends with `return json_decode( $body_response, true )` (`src/Rest_API/Base.php:256`) and `check_response_error()` does not throw on an empty/non-JSON 200 body — so it can return `null`. The original caller tolerates this (`Order\Single::postnl_activate_return_function` reads `! empty( $response['successFulBarcodes'] )` on a possibly-null value, no fatal), but the new `: array` signature would convert the same response into a `TypeError`. Affects `Legacy\Barcode\Client::generate`, `Legacy\Checkout\Client::get_delivery_options`/`get_pickup_locations`, `Legacy\Postcode_Check\Client::check`, `Legacy\Smart_Returns\Client::generate`, and `Barcode_Service`/`Checkout_Service`/`Postcode_Check_Service`/`Smart_Returns_Service`/`Return_Label_Service::activate` — *latent* (nothing calls them until task 9). Fix: normalize before returning (`$res = $client->send_request(); return is_array( $res ) ? $res : array();`) or widen to `?array`, before task 9 wires the callers.
- 🟢 **Low — deployment / autoload fragility** The 16 original-path shims (e.g. `src/Rest_API/Shipping/Client.php`, `src/Rest_API/Barcode/Client.php`) contain only a top-level `class_alias()` with no class/interface declaration. This is **safe today** — `deploy.yml:132` uses only `--optimize-autoloader`, which keeps the PSR-4 fallback that loads the shim and fires the alias. But under `--classmap-authoritative` (`-a`, a common prod-hardening addition) the scanner finds no symbol, the file is dropped from the classmap, the PSR-4 fallback is disabled, and every legacy FQCN fatals with "Class not found". Reproduced. Fix (defensive): make each shim scanner-visible — declare a real subclass in the original namespace that `extends` the Legacy class (and a matching `Item_Info` subclass), or at minimum document/lock that the build must never use `-a`.
- 🟢 **Low — missing test coverage** No test added for the acceptance criterion. `tests/php/integration/PluginBootstrapTest.php` only asserts constants/Utils; PHP enforces `implements` only at class-load, so future interface/client signature drift would pass CI silently. Fix: add a cheap unit test asserting `class_implements()` for each Legacy client/wrapper against its contract, plus an alias smoke test that each old FQCN resolves to its Legacy class.
- ℹ️ **Info — redundancy (by design)** Each raw-data interface is implemented twice (e.g. `Legacy\Barcode\Client::generate` and `Legacy\Barcode_Service::generate` have identical bodies). The Client's instance method ignores `$this` and builds a fresh `self`, so it's only reachable after a Client already exists — which is exactly why the zero-arg wrapper is what the factory injects. Acceptable as-is; the `implements` on the raw Client satisfies the spec. Optional later cleanup: collapse to a single source of truth.
- ℹ️ **Info — pre-existing PHPCS debt (not a regression)** PHPCS reports errors/warnings in moved files (mostly `Legacy/Shipping/Item_Info.php`); counts are identical to the originals on the base branch and the bodies are byte-identical, so this is carried-over debt, not introduced here. The genuinely-new code (wrappers, shims, added methods) is clean. Do not fold a WPCS cleanup into this mechanical-move PR.

**Testing notes**
- Confirm the alias chain in a real PHP process: autoload and `class_exists('PostNLWooCommerce\\Rest_API\\Shipping\\Client')` / `Barcode\\Client` etc. resolve to the `Legacy\*` classes, and `class_implements()` on each Legacy client/wrapper includes its contract interface.
- Exercise the existing label pipeline end-to-end (single + bulk label generation, return-label activation, letterbox) — `Order\Base` imports `Shipping`/`Return_Label`/`Letterbox`/`Barcode` by the OLD namespace, so this proves the shims resolve under the deploy autoloader. Test both classic and blocks checkout for the Checkout/delivery-options path.
- If prod hardening with `composer install ... --classmap-authoritative` is ever on the table, build with `-a` locally first and confirm no "Class not found" on legacy FQCNs (currently would fatal — see 🟢 above).
- Before task 9 wires the new typed methods, simulate an empty/204/non-JSON 200 response and confirm it no longer throws a `TypeError` once the return is normalized.

dustinparker (approved):
Everything is working as expected. The move into `Legacy\` is faithful, with the client and item-info logic carried over unchanged, and the alias shims keep every existing `Rest_API\<Flow>\*` reference resolving, so the live label and checkout paths behave exactly as they did before.

The service wrappers also close the gap I raised on the interfaces PR: `send_request()` returns whatever `json_decode()` gives it, so normalizing to an array is what keeps the `: array` contract honest on an empty or non-JSON response rather than turning a handled error into a fatal.

PR #325, opened 2026-06-06
Postnl v4 service factory

sal4sup:
### All Submissions:

Add `Service_Factory` (Subtask 8) — the single resolver that will later route each
  flow to Legacy V1 or the PostNL V4 SDK.

Closes # .
https://app.clickup.com/t/4550855/868jqaqf9

1.
2.
3.

### Other information:

### Changelog entry


Abdalsalaam:
## 🤖 Automated PR review
**ClickUp task:** [8. Service_Factory (defaults to Legacy)](https://app.clickup.com/t/868jqaqf9)

Verified against the incremental diff (`origin/postnl-v4-legacy-clients-interfaces..origin/postnl-v4-service-factory`): one new factory + its unit test, no other production wiring. Legacy-by-default is correctly implemented and well tested.

**Spec coverage**
- ✅ New `src/Rest_API/Service_Factory.php` as the single Legacy↔V4 decision point — one method per flow.
- ✅ Methods return the flow's interface type (`barcode_service(): Barcode_Service_Interface`, etc.); each Legacy service implements its interface, and `Checkout_Service` implements both `Timeframe_` and `Pickup_Location_` interfaces.
- ✅ V4 gate is correctly dormant: `should_use_v4()` requires key **and** `Router::sdk_enabled_for()`, plus `isset($v4_services[flow])` — with no `inject_v4_service()` call in production, every method falls through to Legacy.
- ✅ Cheap-when-off: `has_v4_key()` short-circuits before Router/filter; Legacy services are built lazily and memoised; nothing is constructed eagerly.
- ✅ Acceptance tests: no-key → Legacy for all 8 methods (plus no-getter, empty, whitespace-key cases); key+flag+stub → returns the V4 stub for every eligible flow; per-flow isolation and `postcode_check` Legacy-only are both asserted.
- ⚠️ "Logic … AND (product-coded) `V4_Mapper::has_v4_equivalent(...)`" — not implemented. `V4_Mapper` does not exist on this branch; the gate is deferred. Acceptable by design but it is a real gap versus the written spec.
- ⚠️ "Inject Settings **and a lazily-built SDK client** (task 10)" — only `Settings` is injected; there is no SDK-client parameter/property/lazy builder. Cheap-construction intent is met, but the named SDK-client injection point is absent.
- ⚠️ `shipment_and_return` is a `Router::SUPPORTED_FLOWS` entry with a Legacy `Shipment_and_Return` client, but has no factory method (no service wrapper/interface exists for it yet).

**Findings**
- 🟢 **Low — spec-deviation** `V4_Mapper::has_v4_equivalent()` gate for product-coded flows is absent in `should_use_v4()` `src/Rest_API/Service_Factory.php:256-258` — confirm it is deliberately deferred (V4_Mapper not on branch) and add a `TODO`/docblock note referencing the dependent task so the gate isn't forgotten when V4 services land.
- 🟢 **Low — spec-deviation** No lazily-built SDK client injected; constructor stores only `$settings` `src/Rest_API/Service_Factory.php:83-85` — confirm with the task owner that the task-10 SDK client wiring lives elsewhere, or thread a lazy SDK-client dependency through now so V4 registration can use it.
- 🟢 **Low — coverage-gap** No `shipment_and_return_service()` method for the `shipment_and_return` SUPPORTED_FLOW `src/Rest_API/Service_Factory.php` — document that S&R is intentionally out of scope here (it has no service/interface yet) or add the interface + method when it lands.
- ℹ️ **Info — dead-flow** The `checkout` flow in `Router::SUPPORTED_FLOWS` is never queried; the factory uses `timeframe`/`pickup_location` keys instead `src/Rest_API/Service_Factory.php:137-157` — so toggling `postnl_sdk_enable_checkout` has no effect here. Worth a one-line note to spare whoever wires V4 later.
- ℹ️ **Info — robustness** `inject_v4_service()`/`set_legacy_service()` accept any flow string without validating against `Router::SUPPORTED_FLOWS` `src/Rest_API/Service_Factory.php:97-99,113-115` — a typo'd flow is silently stored and never read. Optional dev-time guard; low priority for internal/test helpers.

**Testing notes**
- Run `tests/php/unit/Rest_API/Service_FactoryTest.php` (Brain Monkey) — confirm all scenarios green, especially the data-provider V4-stub cases and `test_postcode_check_always_legacy_ignores_v4_injection`.
- Sanity-check that with a real V4 key set but every `postnl_sdk_enable_*` filter at its default `false`, all eight getters still return Legacy (the production state today).
- No checkout-mode (classic vs blocks) or runtime impact: this PR only adds a factory + test and is not yet referenced by any caller — confirm no production code path constructs `Service_Factory` in this diff.
- If/when V4 services are wired, re-verify the product-coded gate (`label`, `letterbox`, `return_label`, `smart_returns`, `barcode`) once `V4_Mapper::has_v4_equivalent()` exists.

dustinparker (approved):
Everything is working as expected. Every flow defaults to Legacy, the postcode check stays pinned there by design, and the shared checkout service is memoised so timeframe and pickup lookups only build it once.

I pushed one fix. `has_v4_key()` was checking for a `get_v4_api_key()` method that no settings class defines and no task specifies, so the V4 gate could never fire even with a key entered and the flag on. It now gates on `get_api_key_new()` and `is_api_key_new_validated()`, which is what the New API Key field actually provides, and it requires the key to have passed validation before routing to V4. The settings double was defining the old method itself, which is why this stayed green. Worth carrying into the V4 Timeframes PR, which has the same check.

PR #326, opened 2026-06-06
Postnl v4 cut callers to factory

sal4sup:
### All Submissions:

* [] Have you tested both blocks/non-blocks cart/checkout?
* [] Have you tested the cart and checkout as both a logged-in user and a guest?

  - `Frontend\Container::get_checkout_data()` and `validated_address()` now call `service_factory()->timeframe_service()->get_delivery_options()` and
  `postcode_check_service()->check()` respectively. One checkout API call is preserved (the full response includes both delivery options and pickup
  locations).
  - `Order\Single::postnl_activate_return_function()` and `postnl_send_smart_return_email()` now call `return_label_service()->activate()` and
  `smart_returns_service()->generate()` respectively.
  - `Order\Base::create_barcode()` and `maybe_create_return_barcode()` now call `barcode_service()->generate()`.
  - `Order\Base::create_label()`, `maybe_create_letterbox()`, and `maybe_create_return_label()` are now thin factory wrappers. Their original bodies have
  been extracted into protected `*_pipeline()` methods (`create_label_pipeline`, `maybe_create_letterbox_pipeline`, `maybe_create_return_label_pipeline`) to
  prevent infinite recursion: `Legacy\Label_Service`, `Legacy\Letterbox_Service`, and `Legacy\Return_Label_Service` extend `Order\Base` and call these
  pipeline methods directly, bypassing the factory.
  - `Order\Base` and `Frontend\Container` each gain a lazy `service_factory()` getter (initialised from the existing `$this->settings` instance).
  `Order\Single` inherits the getter from `Order\Base`.

Closes # .
https://app.clickup.com/t/4550855/868jqaqfb

1. With no V4 API key configured (default state), open a paid NL order in wp-admin, generate a shipping label from the PostNL meta box, and confirm the label PDF, barcode, and `_postnl_order_metadata` are identical to before this refactor (the factory resolves the Legacy client for every flow).
2. Repeat for a letterbox-eligible order and for an order with "Create return label" enabled — letterbox and return labels generate as before.
3. Classic (shortcode) checkout: enter a NL address, verify delivery-day slots and pickup points load, select one of each variant, place the order, and verify the selection is stored on the order.
4. Blocks checkout: repeat step 3.
5. From an order with a generated label, trigger "Activate return" and "Send Smart Return email" and confirm both still work (they now go through `return_label_service()->activate()` / `smart_returns_service()->generate()`).
6. Classic checkout with postcode validation enabled: enter a valid NL postcode/house number and confirm the address autocompletes (postcode check now routes through the factory, permanently Legacy).
7. Confirm no `postnl_sdk_enable_*` filter is enabled anywhere — every flow must resolve to a Legacy service, so runtime behavior is byte-identical to the base branch.

### Other information:

### Changelog entry


Abdalsalaam:
## 🤖 Automated PR review
**ClickUp task:** [9. Cut callers over to the factory](https://app.clickup.com/t/868jqaqfb)

**Spec coverage**
- ✅ **Container.php (timeframe + postcode)** — `get_checkout_data()` now calls `service_factory()->timeframe_service()->get_delivery_options()` and `validated_address()` calls `postcode_check_service()->check()`; zero `new Checkout/Postcode_Check Client` or `send_request()` remain.
- ✅ **Base.php (barcode, label, return, letterbox)** — both `create_barcode` paths route through `barcode_service()->generate()`; `create_label` / `maybe_create_return_label` / `maybe_create_letterbox` are now thin factory wrappers with bodies extracted into protected `*_pipeline()` methods.
- ✅ **Single.php + AJAX handlers** — `postnl_activate_return_function` → `return_label_service()->activate()`, `postnl_send_smart_return_email` → `smart_returns_service()->generate()`; four stale imports removed, no direct instantiation left.
- ✅ **Bulk.php** — not edited, correctly so: its label flow is `generate_label_and_notes()` → `save_meta_value()` → `create_label()`/`maybe_create_return_label()` (now factory wrappers), so it is cut over transitively.
- ✅ **100% Legacy at runtime / byte-identical** — `Settings` has no `get_v4_api_key()` on this branch, so `Service_Factory::has_v4_key()` returns false and every flow resolves to Legacy; `Router::sdk_enabled_for()` also defaults every flow to false. Pipeline bodies are verbatim extractions.
- ✅ **Recursion avoidance** — the three `Order\Base`-derived Legacy services call `*_pipeline()` (not the public wrappers), and the factory returns *new* Legacy service instances, so there is no `create_label()` → factory → service → `create_label()` loop.
- ✅ **Lazy `service_factory()` getter from settings** — Base (protected) and Container (private) each add a null-memoised getter building `new Service_Factory( $this->settings )`.
- ❓ **End-of-phase acceptance ("disable Legacy, every flow errors cleanly")** — manual; not exercisable from this incremental diff.
- ❓ **Manual smoke test (classic + blocks + label gen)** — manual; see testing notes re: blocks path.

**Findings**
- 🟢 **Low — stale-documentation** — All three Legacy service docblocks still say they delegate to the *public* methods (`Label_Service`: "Delegates entirely to `Order\Base::create_label()`"; `Letterbox_Service`: `maybe_create_letterbox()`; `Return_Label_Service`: `maybe_create_return_label()`), but the code now calls the `*_pipeline()` variants. Following the docs literally would re-route through the factory and reintroduce the exact recursion this PR avoids `src/Rest_API/Legacy/Label_Service.php`, `Letterbox_Service.php`, `Return_Label_Service.php` — update each docblock to reference the `*_pipeline()` method and note the recursion-avoidance reason (mirroring the new docblocks on the Base.php pipeline methods).
- 🟢 **Low — missing-test-coverage** — No automated test guards this PR's core invariant: that the Legacy services' `create()` reaches the `*_pipeline()` body without recursing, and that the public wrappers route through `Service_Factory`. The carried-over `Service_FactoryTest.php` only exercises routing via injected stubs `tests/php/unit/Rest_API/` — add a test that a real Legacy_*_Service `create()` reaches the pipeline (e.g. spy/assert no re-entry) and that the public wrappers delegate to the factory; if a WC/Settings harness is impractical, note it in the manual-test section.

**Testing notes**
1. **Classic checkout** is directly cut over here — verify delivery options load and the postcode/address validation still work on a NL address (`Container::get_checkout_data` / `validated_address`).
2. **Blocks checkout** delivery options flow through `Extend_Block_Core`, which is **not** touched in this diff — confirm that path was cut over in a prior stacked PR or is out of scope, then smoke-test blocks delivery options + pickup points.
3. **Label generation** — generate a single-order label and a bulk label; confirm output (barcode, PDF, merged sheet) is byte-identical to pre-PR behavior, since Bulk relies entirely on the inherited Base wrappers.
4. **Return + smart return** — activate a return label (`postnl_activate_return_function`) and send a Smart Return email (`postnl_send_smart_return_email`); the `(int) $order_id` cast on activate is behaviorally equivalent for the WC order lookup.
5. **Recursion sanity** — exercise letterbox/return label creation end-to-end with logging on to confirm no infinite-recursion / stack-depth issue surfaces at runtime.

dustinparker (approved):
Everything is working as expected. The cut-over is faithful: every caller resolves its service through the factory now, and the barcode, label, checkout and postcode paths return exactly what they returned before, with the response normalisation leaving the existing `empty()` guards behaving identically. The shared checkout service still hands back delivery options and pickup locations from the one call, so the frontend keeps both.

One thing worth flagging for whoever touches these next. The wrappers extend `Order\Base` and inherit its public label methods, so each `create()` has to call the protected `*_pipeline()` variant. Calling the public method instead sends every hop back through the factory into a fresh wrapper and exhausts memory, which is a fatal on the label path rather than a handled error, and the trap invites the mistake because `create_label()` is the obvious method while `create_label_pipeline()` is the odd one. I pushed integration tests that pin the invariant for all three wrappers, so a regression fails as a readable assertion naming the method to call instead of an out-of-memory crash that takes the suite down with it.

PR #327, opened 2026-06-06
Postnl v4 sdk client factory

sal4sup:
### All Submissions:

Closes # .
https://app.clickup.com/t/4550855/868jqaqfu

1.
2.
3.

### Other information:

### Changelog entry


Abdalsalaam:
## 🔍 Structured review — V4 SDK `Client_Factory` (Task [#10 / 868jqaqfu](https://app.clickup.com/t/4550855/868jqaqfu))

Dual-engine review + **adversarial verification** (every finding re-checked against the actual code; refuted candidates listed at the bottom). CI is green and the acceptance criteria are all met.

**Context lens:** this is a **flag-off, no-caller scaffolding PR** (`Client_Factory` is correctly unwired — verified repo-wide; lazy PSR-4; `postnl_sdk_enable_*` default off). **None of the below blocks merging this increment.** The 🟠/🟡 items are **prerequisites before any V4 flow is enabled** (wiring/rollout) and are flagged as such.

> ℹ️ Branch isn't rebased, so `git diff main..HEAD` also shows prior merged PRs (#300/#315/#316). The Task-10 commit is `d453d3c` (composer.json, composer.lock, `src/Rest_API/SDK/Client_Factory.php`, its test).

---

### 🟠 High — resolve before wiring a flow ON

**H1 — PHP 8.2 SDK floor unreconciled with the plugin's `Requires PHP: 7.4`.**
`composer.json` puts `postnl/api-client-sdk` in `require` (prod); its constraint in `composer.lock` is `"php": ">=8.2 <8.5"`. There is **no `config.platform.php`** and `postnl-for-woocommerce.php:12` still says `Requires PHP: 7.4`. CI is green only because deploy.yml / `.wp-env.json` force PHP 8.2. Consequences: (a) contributors on PHP 7.4–8.1 get an unsatisfiable `composer install`; (b) the shipped 8.2-syntax SDK is autoload-registered on every site but PSR-4-lazy, so safety rests on *"no code references an SDK class"* — **the default-off flag gates behavior, not autoloading**, so one wiring mistake on a sub-8.2 site = parse-time fatal the flag won't prevent.
**Fix (wiring task):** gate SDK use on `PHP_VERSION_ID >= 80200` in addition to key+flag; decide whether to bump `Requires PHP` (header + readme.txt) or keep 7.4 with a hard runtime guard; add `config.platform.php`; fix the misleading `deploy.yml:85` comment ("needs to match minimum plugin requirement" — the minimum is 7.4, not 8.2).

**H2 — Unscoped vendor deps → WordPress cross-plugin class collision; the flag does NOT contain this.**
The SDK pulls **Guzzle, league/oauth2-client, psr/\*, symfony polyfills** into prod `vendor/`, autoloaded **unprefixed** (no Strauss/PHP-Scoper). `postnl-for-woocommerce.php:50` requires `vendor/autoload.php` unconditionally at boot, and four packages use Composer `files` autoload (guzzle `functions_include.php`, `ralouphie/getallheaders`, symfony `bootstrap.php`/`function.php`) that **eager-load on every request regardless of the flag**. WP shares one namespace across all active plugins; "first autoloader wins" can bind PostNL's SDK to another plugin's skewed Guzzle. (Guzzle is in prod only because it satisfies the SDK's virtual `psr/http-client-implementation` — scoping must cover whatever fulfills that.)
**Fix (before rollout):** scope runtime deps under a plugin-private prefix (Strauss) as a build step between `composer install --no-dev` and packaging; add a CI gate asserting no unprefixed `GuzzleHttp\`/`Psr\Http\` in the shipped `vendor/composer/autoload_psr4.php`.

---

### 🟡 Medium

**M1 — Two live distribution paths, not in lockstep.** Deploy ships via `rsync --exclude-from=.distignore`; `npm run build` → `npm run archive` → **`composer archive`** (`package.json:8,27`) ships via **`composer.json` `archive.exclude`** (this PR edited it, adding `auth.json` — so it's live, not dead config). Both exclude auth files and ship the SDK today (**no leak**), but editing one path silently desyncs the other. Both also ship the SDK's bundled dev tree (`vendor/postnl/api-client-sdk/tests`, `/phpstan`, `/scripts`, CS/rector/phpstan configs).
**Fix:** keep `.distignore` and `archive.exclude` in lockstep (or comment which governs which); strip the SDK's non-runtime files from both (ideally `.gitattributes export-ignore` upstream).

**M2 — `declare(strict_types=1)` + untyped Settings getters = latent `TypeError`; tests mask it.** `build()` (`Client_Factory.php:78-79`) feeds `Settings::get_customer_num()/get_customer_code()` (untyped, no string guarantee) into `make_builder(string …)` and concatenates them into the cache key. Corrupted/legacy option data (null/array/false) warns at the concat and throws at the typed call. The unit stub types its getters `: string` — **stricter than real `Settings`** — so the suite is blind to this.
**Fix:** cast at the boundary — `$customer_number = (string) $this->settings->get_customer_num();` (and code) — and drop `: string` on the stub + add a non-string case.

**M3 — Test coverage gaps on shipped behavior.** (a) `withRetry(RetryConfig::exponentialBackoff())` (`Client_Factory.php:116`) is configured but **never asserted**. (b) Customer creds are in the cache key but **no test proves two different-credential settings yield different clients** — the `customer_*` segments are uncovered. (c) Config is asserted by reflecting the **third-party SDK's private `ClientBuilder` props** on a `dev-main`-pinned package — brittle; the cleaner seam (capturing `make_builder()` args via `Spy_Client_Factory`) already exists.
**Fix:** assert retry config; add a credential-discrimination test; prefer asserting captured `make_builder()` args over SDK-private reflection.

**M4 — `php-tests.yml` runs on fork PRs without secret access.** Trigger is bare `on: pull_request` with no fork guard; GitHub withholds `secrets.POSTNL_PRIVATE_PACKAGIST_AUTH` from forks → `composer install` 401s on the private mirror → whole job **red** for external contributors, breaking "every PR green." Fine **iff** the repo never takes fork PRs.
**Fix:** confirm internal-only (and comment it), or guard the auth step with `if: github.event.pull_request.head.repo.full_name == github.repository` + a no-SDK fallback for forks.

---

### 🟢 Low

- **L1** — SDK pinned to an unpinned dev branch (`composer.json` `"1.x-dev@dev"` → lock `dev-main`, commit-pinned so installs are reproducible; all dist `shasum` empty is a lockfile-wide Private-Packagist trait). Move to a semver tag (`^1.0`) before rollout. *(The SSH-vs-HTTPS "fresh resolve breaks" worry is refuted — `--prefer-dist` + HTTPS mirror; CI green proves it.)*
- **L2** — `test_no_message_id_referenced` (`Client_FactoryTest.php:174-180`) is a source-grep near-tautology coupled to the unit bootstrap's plugin-root `ABSPATH`; wouldn't catch a MessageID added by the SDK or elsewhere. Make it behavioral via the spy, or mark it explicitly as a source-lint.
- **L3** — `php-tests.yml` pins no PHP (no `setup-php`); green depends on the runner default (8.3) staying inside `[8.2,8.5)`. Add `shivammathur/setup-php@v2` with `8.2`.
- **L4** — Bare `object` constructor hint (`Client_Factory.php:62`) discards the `get_customer_*` contract (runtime-only). Optional: a small `interface` both `Settings` and the stub implement.
- **L5** — Add a memo-key guard comment at `Client_Factory.php:81`: any future `with*()` sourced from settings/request data must be added to the cache key or memoization returns a misconfigured client. (Key is complete today.)
- **L6** — SDK dev tree shipped to WP.org (footprint; see M1).

---

### ℹ️ Informational / checked

- `composer test` runs `composer config name`, **not** PHPUnit — footgun; real entrypoints are `npm run test:php:unit|integration`.
- **No `phpcs`/`check-php` job in any workflow** → the declared `testVersion 7.4-` compat sniff for `src/` is never enforced in CI.
- Release version-gate (`deploy.yml` validate) checks only the PHP header + `readme.txt` — not `src/Main.php` or `package.json` (agents.md says all four must match). Pre-existing infra; worth hardening as the migration adds bump surfaces.
- `@since 5.9.6` is premature (current is 5.9.5) — trivial.
- **Confirmed correct ✅:** secret handling is clean (token only via `COMPOSER_AUTH` env, never echoed; auth-leak + ZIP gates fail-closed); license MIT ↔ GPL-3.0 compatible; `--no-dev` reinstall before packaging means **dev deps do NOT ship**; no-caller / lazy-PSR-4 / flag-default-off invariant holds (integration smoke test boots the whole plugin touching no SDK class). Key-hashed memo cache key is a nice security touch.

**Refuted candidates (checked, not issues):** dev-deps-in-ZIP (a `--no-dev` reinstall exists before packaging); `$sut` naming convention (none exists in this repo); `@testdox` trailing-period "lint" (tests are phpcs-excluded; no such rule); per-method `@since` "requirement" (used 10× across 395 methods); `find -maxdepth` leak gap (theoretical — the ZIP scan is unbounded).

---

### Testing / build notes
- **Local build & SDK unit tests are not reproducible locally** — the private SDK is auth-gated (`vendor/postnl/api-client-sdk` empty without `COMPOSER_AUTH`); CI is the only environment that exercises them (✅ green). Recommend linking the green CI run in the PR description.
- `php -l` passes on both new files; `composer validate` clean (its two warnings — deprecated `GPL-3.0` SPDX id, missing `description` — are pre-existing).
- Suggested regression check before any rollout: with all flags off, place a legacy order + load checkout delivery options (classic **and** blocks) and confirm parity with `main`, and that no `Postnl\Sdk\*` class is autoloaded on a normal request.

<sub>🤖 Automated structured review (dual-engine + adversarial verification). Findings are advisory; the maintainer decides what to action.</sub>

dustinparker (approved):
Everything is working as expected. The factory builds a correctly configured V4 client with SourceSystem 35, the sandbox flag, and the customer credentials all reaching the SDK, nothing goes over the wire at construction, and MessageID stays off the V4 path.

I pushed `ed8c237` with three fixes. The memo cache key joined the customer values around an unescaped `|`, so `('1|2', '3')` and `('1', '2|3')` produced the same key and collided onto one client, handing the second configuration the first one's credentials. The customer values are country-scoped and can differ within a request, so each segment is now hashed to a fixed width. The raw V4 key was also readable in exception stack traces despite the docblock claiming the opposite, so it is marked `SensitiveParameter` on `build()` and `make_builder()`, on its own line so it stays a comment on the declared 7.4 floor. And `@since` pointed at 5.9.6, which already shipped, so it now matches `Service_Factory` at 5.9.9.

The tests needed the same attention. Swapping `Auth::apiKey( $v4_key )` for a hardcoded literal left the whole suite green, so the key forwarding is now asserted, and the no-network test only checked the return type, which holds whether or not a request goes out, so it now injects a PSR-18 spy instead.

One thing for when this gets wired up in the next tasks: the memo is per-instance, so it only pays off if `Client_Factory` ends up a shared service rather than being constructed per call.

PR #328, opened 2026-06-23
Remove: Discontinued "Shipment & Return"

sal4sup:
### All Submissions:

PostNL is stopping the "Shipment & Return" product on 1 July 2026. This PR removes it from the plugin.
- Removed "Shipment & Return" from the Standard return option dropdown in settings. Now only "None" and "Label in the box" are left.
- Shops that still had "Shipment & Return" selected are changed to "None" automatically when they update. This runs one time only.
- This stops orders from failing after 1 July, when PostNL turns the product off.

Closes # .
https://app.clickup.com/t/868jxyjp1

1. Before updating, set the Standard return option to "Shipment & Return" (on the old version) and save.
2. Update to this version and open the PostNL settings again.
3. The setting is now "None", and "Shipment & Return" is gone from the dropdown. Only "None" and "Label in the box" show.
4. Place a test order and make a label — it works with no "Shipment & Return" errors.
  
### Other information:

### Changelog entry

> Remove: Removed the Shipment & Return return option, as PostNL is discontinuing it on 1 July 2026.


dustinparker (approved):
Everything is working as expected. The Shipment & Return option is removed from the Standard return option settings, any store still set to it is reverted to None on update, and a lingering value is treated as None at read time, so no outbound label can request the product PostNL discontinues on 1 July 2026. None and Label in the box still behave exactly as before.

PR #330, opened 2026-06-29
Postnl v4 sdk logger adapter

sal4sup:
### All Submissions:

* [x] Does your code follow WooCommerce and WordPress standards? (PHPCS clean)
* [x] Have you successfully run tests with your changes locally? (full unit suite green)
* [ ] Tested both blocks/non-blocks checkout — N/A: no checkout/runtime change (flag off, adapter not yet wired into a caller)
* [ ] Tested as logged-in user and guest — N/A
* [ ] Placed an order as logged-in user and guest — N/A
* [ ] Query Monitor active during testing — N/A

Task 11 — PSR-3 logger adapter for the V4 SDK. Stacked on #327 (task 10).

Adds `Rest_API\SDK\Logger_Adapter` (extends `Psr\Log\AbstractLogger`) to bridge the SDK's PSR-3 logging to the plugin's `WC_Logger`:

* Tags V4 messages `[postnl-v4]` on the shared `PostNLWooCommerce` log source so support can filter by API version.
* Maps PSR-3 levels 1:1 to WC levels (unknown → `notice`); interpolates `{placeholder}` context.
* Honors the existing "enable logging" toggle via `Logger::is_enabled()`; relies on the SDK's `RedactionRegistry::forProduction()` for redaction — no `check_pdf_content()` on the V4 path. `src/Logger.php` is unchanged.

No behavior change: the adapter is not yet wired into `Client_Factory`.

ClickUp: https://app.clickup.com/t/868jqaqfy

1. `composer install`, then run `vendor/bin/phpunit -c phpunit-unit.xml.dist --filter Logger_Adapter` (or `npm run test:php:unit`).
2. Confirm `Logger_AdapterTest` passes — covers tagging, level mapping, interpolation, unknown-level fallback, and disabled-no-write.

### Other information:

* [x] Checked there aren't other open PRs for the same change.

### Changelog entry

Add V4 SDK PSR-3 logger adapter (`[postnl-v4]`-tagged) bridging SDK logging to WC_Logger. No behavior change; feature flag off.


dustinparker (approved):
Everything is working as expected. The SDK's redaction holds through the adapter, so the API key and label binary never reach the WC log, and the `[postnl-v4]` tag keeps V4 entries filterable on the shared source alongside the legacy ones. I pushed a commit forwarding the PSR-3 context to `WC_Logger`, which keeps entries complete under either SDK log format, and corrected the `@since` to 5.9.9.

PR #331, opened 2026-06-29
Task 12: WP-transient PSR-16 cache adapter for V4 SDK (timeframe/loca…

sal4sup:
### All Submissions:

* [x] Does your code follow WooCommerce and WordPress standards? (PHPCS clean under .phpcs.xml)
* [x] Have you successfully run tests with your changes locally? (full unit suite green)
* [ ] Tested both blocks/non-blocks checkout — N/A: no checkout/runtime change (flag off, adapter not yet wired into a caller)
* [ ] Tested as logged-in user and guest — N/A
* [ ] Placed an order as logged-in user and guest — N/A
* [ ] Query Monitor active during testing — N/A

Task 12 — WP-transient cache adapter for the V4 SDK. Stacked on #327 (task 10).

Adds `Rest_API\SDK\Cache_Adapter`, extending the SDK's `AbstractCacheAdapter` (implements `CacheAdapterInterface` = PSR-16 + `isAvailable()`), backed by WordPress transients:

* Caches only allowlisted keys (`timeframe`/`locations`, filterable via `postnl_v4_cache_allowed_prefixes`); any other key bypasses the cache.
* TTL defaults to 600s, filterable via `postnl_v4_cache_ttl`; transient keys are namespaced with a hash of the V4 key (cross-tenant isolation); `clear()` is scoped to that namespace.

Notes (SDK differs from the original task wording): the SDK has no "CachingPlugin"/response cache — its cache contract is `CacheAdapterInterface`, not bare PSR-16, and keys are sanitized (no `/`), so the allowlist is enforced by key prefix. This adapter is what tasks 15/16 (timeframe/locations + caching) will use. No behavior change: not yet wired into a caller.

ClickUp: https://app.clickup.com/t/868jqaqg3

1. `composer install`, then run `vendor/bin/phpunit -c phpunit-unit.xml.dist --filter Cache_Adapter` (or `npm run test:php:unit`).
2. Confirm `Cache_AdapterTest` passes — covers transient round-trip, allowlist bypass (get & set), filterable TTL + fallback, per-tenant namespacing, invalid-key exception, and scoped `clear()`.

### Other information:

* [x] Checked there aren't other open PRs for the same change.

### Changelog entry

Add V4 SDK WP-transient cache adapter (timeframe/locations allowlist, filterable TTL). No behavior change; feature flag off.


dustinparker (approved):
Everything is working as expected. Only allowlisted keys are cached, a zero or negative TTL falls back to the default instead of caching permanently, and clear() reports failure rather than a success that cleared nothing.

PR #335, opened 2026-07-08
Task 13: SDK exception converter (V4 SDK errors -> legacy error shape)

sal4sup:
### All Submissions:

* [ ] Have you tested both blocks/non-blocks cart/checkout? — N/A: backend-only plumbing, nothing invokes it (V4 flag off).
* [ ] Have you tested the cart and checkout as both a logged-in user and a guest? — N/A.
* [ ] Have you successfully placed an order as both a logged-in user and a guest? — N/A.
* [ ] Did you have the query monitor plugin active during all testing? — N/A.

Task 13 of the V4 SDK migration. Adds `Rest_API\SDK\Exception_Converter` — a static `convert( \Throwable ): \Exception` that translates V4 SDK exceptions into the exact error shape the legacy REST clients throw (human-readable `getMessage()` + HTTP status in `getCode()`), so `Order\Base` and the frontend handlers need no special-casing for the V4 path.

Mapping:
- Auth failures (HTTP 401/403 and pre-request credential/OAuth) → "Invalid PostNL API credentials".
- Validation (400/422) → bubbled `field: message` list (falls back to the `ProblemDetails` message when no field errors).
- Transient failures (429, 408, 5xx, network, retries exhausted) → "PostNL temporarily unavailable, please try again."
- Any other HTTP error → parsed `ProblemDetails` message.
- Non-HTTP SDK/runtime errors → own message passed through.

The PostNL `traceId` is appended for support correlation, the status code is preserved via `getCode()`, and the original SDK exception is kept as `previous` so the full cause chain stays available for logging. No behavior change for merchants — flag stays off; nothing calls this yet.

ClickUp: [13. SDK exception converter](https://app.clickup.com/t/868jqaqg5) (depends on task 10).

> **Stacked PR:** based on `postnl-v4-sdk-client-factory` (task 10, #327), matching #330/#331. The diff will show task 10's files until that PR merges; review [the two-file range](../compare/postnl-v4-sdk-client-factory...postnl-v4-sdk-exception-converter) for this task's changes only.

1. `composer install` (pulls the V4 SDK from private packagist).
2. Run the unit suite: `./vendor/bin/phpunit -c phpunit-unit.xml.dist --filter Exception_Converter` — 15 tests, one per SDK exception class, asserting the mapped message + preserved status code.
3. `composer check-php` (PHPCS) — passes clean.

### Other information:

### Changelog entry

No changelog entry — internal V4 SDK plumbing behind an off-by-default flag, no merchant-facing change (consistent with #330/#331).


dustinparker (approved):
Everything is working as expected. The converter maps every SDK exception into the shape the existing handlers already read, so the V4 path drops in behind `Order\Base` and the frontend handlers without special-casing anything.

I pushed one commit with two fixes.

Permanent 5xx responses were collapsing into the temporary-failure message. `ServerException` implements `RetryableExceptionInterface` for every 5xx, but `isRetryable()` returns false for the permanent ones, so a 501 told the merchant to retry a call that can never succeed and threw away the explanation PostNL sent with it. The check now asks `isRetryable()` instead of testing the interface, which is the same gate the SDK's own retry policy uses.

The other one is the fallthrough. It passed SDK and PHP internals straight into merchant-facing notices, and a transport failure that normalizes to `LogicSdkException` printed the full request URI including the customer code and number. `HttpSdkException` is now the boundary for passing a message through, and anything past it gets a generic message with the original kept as the previous exception so nothing is lost for logging. That also lines up with how the SDK's own error handling docs split these types.

Two smaller things came along with it. `esc_html__()` became `__()`, since these are message data rather than output and escaping at construction turned an apostrophe in a translation into `&#039;` in the JSON error response. The `@since` tags now point at 6.0.0.

One thing I left alone: the admin JS drops the error message into `error_cont.html()` in four places, so API-derived text still reaches an HTML sink unescaped. That is pre-existing and shared with the legacy path, so it belongs in its own issue rather than here.

PR #336, opened 2026-07-12
V4 barcode ordering — no standalone barcode call

sal4sup:
### All Submissions:

ClickUp: https://app.clickup.com/t/868jqaqg7 — *14. V4 barcode ordering — no standalone barcode call*

PostNL confirmed (2026-05-21) the Barcode API is no longer needed on V4: the label call (`labelconfirm`) auto-issues the barcode and returns it, so there is **no standalone V4 barcode endpoint**. This adapts the barcode flow for V4 while leaving the Legacy (V1) path byte-identical.

- **`src/Rest_API/V4/Barcode/Service.php` (new)** — no-op V4 barcode service implementing `Barcode_Service_Interface`. `generate()` performs no request, so the V4 path never calls `GET /barcode`.
- **`src/Rest_API/Service_Factory.php`** — new `barcode_from_label()` predicate, gated on a **real V4 label service being registered** via `inject_v4_service()` (not merely the barcode flag), so a Legacy label service is never asked to build a request without a prefetched barcode. `barcode_service()` itself is unchanged: the V4 no-op is returned only when explicitly injected — there is no self-wiring.
- **`src/Order/Base.php`** — `save_meta_value()` branches on transport: on V4 it generates the label first and harvests the barcode(s) from the label response (`labels[].barcode`, the existing contract) into the same `barcodes[]` shape; the Legacy branch is unchanged. Adds the pure `get_barcodes_from_labels()` helper.
- **Tests** — V4 no-op service; factory self-wiring + `barcode_from_label()` gate (incl. the safety case); harvest helper.

This is a stacked PR — base branch is `postnl-v4-cut-callers-to-factory`. Everything is behind flags that default off, so trunk stays green with zero behavior change for merchants. Depends on task 18 (V4 Label service) for the end-to-end barcode round-trip; the reorder stays dormant until that lands.

1. Run the unit suite: `vendor/bin/phpunit -c phpunit-unit.xml.dist` — all green (incl. `Service_Factory`, `V4\Barcode\Service`, `Order\Base` harvest tests).
2. Confirm no behavior change on Legacy: with no V4 key, `Service_Factory::barcode_service()` still returns the Legacy service and `barcode_from_label()` is `false`, so `save_meta_value()` runs the unchanged prefetch-then-label path.
3. Confirm the V4 reorder is dormant until task 18: with a V4 key + label flag on but no V4 label service registered, `barcode_from_label()` returns `false` (safety gate).

### Other information:

### Changelog entry

> Dev: adapt the barcode flow for the V4 SDK so the barcode is issued by the label response instead of a standalone barcode call (behind an off-by-default feature flag; no change for existing sites).


dustinparker (approved):
Everything is working as expected. The V4 path now takes the barcode off the label response instead of a standalone request that V4 has no endpoint for, and the reorder is gated on the same condition as the label service, so the Legacy prefetch flow is untouched. A response without a barcode discards the label files rather than leaving orphaned PDFs behind a retry, and the parent barcode fallback keeps the tracking URL populated on merged labels.

PR #337, opened 2026-07-12
15. V4 Timeframes + caching

sal4sup:
### All Submissions:

ClickUp: **15. V4 Timeframes + caching** — https://app.clickup.com/t/868jqaqgd

Adds the V4 SDK-backed delivery-day (timeframe) lookup, mirroring the timeframe half of `Legacy\Checkout\{Client,Item_Info}` behind the `postnl_sdk_enable_timeframe` flag (default off). Nothing runs the new API unless a site has a V4 key **and** the flag is on, so trunk behaviour is unchanged for merchants.

* New `src/Rest_API/V4/Timeframe/Service.php` implementing `Timeframe_Service_Interface`:
  * Builds a `MultipleServicesTimeframeRequest` from the checkout address (`Address_Utils::set_post_data_address()` resolves the billing→shipping fallback, as the legacy path does) plus shipping settings. Daytime is always requested; Evening is added when enabled; morning stays a daytime sub-window (V4 has no separate morning service).
  * `numberOfDays` is a constructor parameter defaulting to **10** (the current plugin value) and clamped to the V4 range `[1, 14]` — no hardcoded cap.
  * Routes through the SDK `CachingPlugin` (backed by task 12's `Cache_Adapter` / WP transients) with **only `/timeframe/` allowlisted**; the key prefix keeps the `Cache_Adapter` allowlist aligned.
  * Maps the V4 response back into the exact `DeliveryOptions` shape `Frontend\Container` consumes (`DeliveryDate` + `Timeframe[]` of `From`/`To`/`Options`), so callers cannot tell V4 from the legacy `/shipment/v1/checkout` path. SDK errors are normalised via task 13's `Exception_Converter`.
* `src/Rest_API/SDK/Client_Factory.php`: adds `build_with_plugins()` so a flow can attach transport plugins (here, the timeframe cache) while reusing the shared auth/version/customer/retry configuration.
* Unit tests (`tests/php/unit/Rest_API/V4/Timeframe/ServiceTest.php`): request built (address, services, clamped `numberOfDays`, billing fallback), response mapped to the legacy shape (incl. morning/evening rules), and a **cache hit on the second identical call** driven through the real `CachingPlugin` + `Cache_Adapter` with a call-counting HTTP client.

Not wired into `Service_Factory`/`Main` yet — consistent with the V4 barcode step; production wiring lands with the flag flip in a later task.

**Stacking:** depends on tasks 8, 10, 12, 13. This branch is built on `postnl-v4-cut-callers-to-factory` (Contracts + Service_Factory + Router + Legacy) with the SDK stack (`Client_Factory`, `Exception_Converter`, `Cache_Adapter`) merged in. Rebase onto `main` once those land so the diff reduces to the timeframe files.

**Handover date:** V4's timeframe request has no `ShippingDuration`/`CutOffTimes` fields (unlike V1), so the calendar walk PostNL used to do server-side now happens in `get_handover_date()`: an order after the cut-off time hands over the next day, each transit day beyond the first adds a preparation day, and the handover lands on the next enabled drop-off day (unit-tested). Parity of the resulting slots should still be confirmed against the PostNL sandbox before the flag is flipped.

**Rollout notes:**
- Do **not** flip `postnl_sdk_enable_timeframe` before the Pickup Locations (task 16) and Checkout aggregation (task 17) work lands: `Frontend\Dropoff_Points` reads `PickupOptions` from the same `get_checkout_data()` response, which the V4 service no longer carries — pickup points would silently disappear.
- Nothing injects this service into `Service_Factory` yet (`inject_v4_service("timeframe", …)`), and `number_delivery_days` from settings still needs to be passed to the constructor at wiring time — the flag-flip task needs a wiring step, not just the filter.

1. `composer install && npm ci`
2. Run the unit suite: `php vendor/bin/phpunit -c phpunit-unit.xml.dist --filter ServiceTest` — 7 tests pass (request building, response mapping, cache hit).
3. Full suite stays green: `php vendor/bin/phpunit -c phpunit-unit.xml.dist` — 144 tests.
4. Coding standards: `composer check-php` — clean.
5. Confirm no behaviour change: with the flag off (default), `Service_Factory::timeframe_service()` still returns the Legacy `Checkout_Service`, so classic and blocks checkout are unaffected.

### Other information:

### Changelog entry

None — internal V4 migration step behind a default-off feature flag; no merchant-facing change.


src/Rest_API/V4/Timeframe/Service.php
```diff
@@ -0,0 +1,482 @@
...
+	private $number_of_days;
+
+	/**
+	 * Service constructor.
+	 *
+	 * @param Client_Factory $client_factory SDK client factory.
+	 * @param object         $settings       Plugin settings instance.
+	 * @param int            $number_of_days Look-ahead days (default 10, capped at the V4 max of 14).
+	 */
+	public function __construct( Client_Factory $client_factory, object $settings, int $number_of_days = self::DEFAULT_DELIVERY_DAYS ) {
```
tommyshellberg: Two problems with this signature.

**1. `$number_of_days` has a default, so forgetting to pass it won't break anything.**

The description says `number_delivery_days` still needs passing at wiring time. Agreed. But if that gets missed, `$number_of_days` quietly falls back to 10. Tests pass, `composer check-php` passes, and a merchant who set 5 delivery days gets 10. Nobody finds out until someone complains about wrong dates.

`Settings::get_number_delivery_days()` already exists (`Settings.php:1191`) and legacy sends it as `Days` (`Base_Info.php:102`).

Drop the default and PHP throws an `ArgumentCountError` the first time the wiring code runs:

```php
public function __construct( Client_Factory $client_factory, Settings $settings, int $number_of_days ) {
```

**2. `$settings` is typed `object` instead of `Settings`.**

That's why you need the `method_exists()` calls below, and it's why a wrong method name can't be caught by anything. Five of the six names you guard definitely exist:

- `get_cut_off_time` - `Settings.php:1284`
- `get_transit_time` - `Settings.php:1275`
- `get_dropoff_days` - `Settings.php:1419`
- `is_evening_delivery_enabled` - `Settings.php:1230`
- `is_morning_delivery_enabled` - `Settings.php:1257`

Those five guards can go. The sixth one is a real problem, see the other comment.

src/Rest_API/V4/Timeframe/Service.php
```diff
@@ -0,0 +1,482 @@
...
+	/**
+	 * Read the V4 API key from settings, tolerating settings without the getter.
+	 *
+	 * The V4 key field ships in a separate PR, so Service_Factory only routes here
+	 * once it exists; the method_exists guard keeps this class safe in isolation.
+	 *
+	 * @return string
+	 */
+	private function get_v4_key(): string {
+		if ( ! method_exists( $this->settings, 'get_v4_api_key' ) ) {
```
tommyshellberg: `get_v4_api_key()` doesn't exist. This always returns `''`.

I'm not arguing with the `method_exists()` approach for getters that land in #309. The problem is the name. Here's every API key method on `origin/feature/new-api-key-field`:

```
828: get_api_key()
837: get_api_key_sandbox()
854: get_api_key_new()              <- the V4 one
864: is_api_key_new_validated()
878: is_api_key_new_validated_value( $key )
904: set_api_key_new_validated( $validated, $key = null )
926: get_effective_api_key()
```

No `get_v4_api_key`. Not on that branch, not anywhere in `src/`. `Service_Factory::has_v4_key()` uses `get_api_key_new()`, which is the right one.

So this guard stays false forever, even after #309 merges and a merchant saves a valid key. Two things break:

1. The SDK client gets built with an empty key, so every request fails auth. The 401 turns into "Invalid PostNL API credentials." That tells the merchant their key is wrong when we never read it.
2. `new Cache_Adapter( '' )` makes the transient prefix `postnl_v4_da39a3ee_` on every site. `Cache_Adapter` hashes the key so two stores don't share cache entries. That stops working.

The tests can't catch it because the settings doubles at `ServiceTest.php:62` and `:256` define `get_v4_api_key()` themselves. They test against a method production doesn't have.

Easiest fix is to pass the key in:

```php
public function __construct( Client_Factory $client_factory, Settings $settings, string $v4_key, int $number_of_days )
```

`Service_Factory::has_v4_key()` has already checked there's a valid key before it would build this, so there's nothing left to guard. If you'd rather read from settings, just rename it to `get_api_key_new()`.

Same call is on `postnl-v4-pickup-locations` and `postnl-v4-checkout-aggregation`, so one fix plus a rebase covers all three.

tommyshellberg (commented):
Reviewed against `main` rather than the PR's base. `git diff origin/main...origin/postnl-v4-timeframes` is 3 files, +1125. The 12-file view is the stale base you mentioned.

The shape is good and I'd merge this structure. Service isn't wired, flag is off, legacy path is untouched. `build_with_plugins()` only adds, it doesn't touch `$this->memo` or change `build()`. The rollout notes are better than most, especially the one about pickup points disappearing if the flag flips before tasks 16 and 17.

Two comments below. They're the same underlying thing. The work you deferred to wiring time is written down in the description, but the code has defaults that make it run anyway. So if the wiring step gets missed, nothing breaks visibly, it just does the wrong thing quietly. Putting those in the constructor signature instead is about five lines, and then the description doesn't have to be remembered.

Left this as a comment rather than requesting changes, but the `get_v4_api_key()` one should get fixed before merge. It stays broken after #309 lands.

Minor: `@since 5.9.6` on the new class is a version that already shipped (we're on 5.9.10), and `postnl_v4_cache_ttl` is documented as `@since 6.0.0` in `Cache_Adapter` but `5.9.6` here.

tommyshellberg (commented):
Second pass with the rest of what I found. These are behaviour and test issues.

Roughly in order of importance:

1. Drop-off days only apply to the first handover, not the rest of the window. Customers can pick dates the merchant can't ship. (`Service.php:322`)
2. Morning and evening are gated differently in `map_option()`. Evening shows up when it's off, and morning gets preselected when it's off. (line 285)
3. Nothing on the V4 path logs anything, and `Logger_Adapter` is written but never used. (line 137)
4. `#[\SensitiveParameter]` missing on `build_with_plugins()`. (`Client_Factory.php:131`)
5. Some tests pass regardless of what the code does. Mutation results on `ServiceTest.php:346`.

Two things I checked that are fine. The cache can't serve one customer's delivery options to another: `CachingPlugin::buildCacheKey()` hashes method + URI + body, and the address is in the body, so different addresses get different keys. And the composer setup is right, SDK is in `require-dev` pinned to `dev-main#5a064dd` with `php >= 7.4` still in `require`, so 7.4 sites are safe.

src/Rest_API/V4/Timeframe/Service.php
```diff
@@ -0,0 +1,482 @@
...
+		if ( $now->format( 'H:i' ) > $this->get_cut_off_time() ) {
+			$handover = $handover->modify( '+1 day' );
+		}
+
+		$extra_days = $this->get_shipping_duration() - 1;
+		if ( $extra_days > 0 ) {
+			$handover = $handover->modify( '+' . $extra_days . ' days' );
+		}
+
+		$dropoff_days = $this->get_dropoff_days();
```
tommyshellberg, line 435: This only moves the first handover to a valid drop-off day. The rest of the window ignores the merchant's drop-off days completely.

Legacy didn't work that way. `Legacy\Checkout\Client::get_cutoff_times()` (lines 89-95) sends one entry per excluded day:

```php
foreach ( $excl_dropoff_days as $day ) {
    $cutoff[] = array( 'Day' => $day, 'Available' => false );
}
```

PostNL uses that to hide any delivery date that would need a handover on a day the merchant doesn't ship.

Here's what happens now. Merchant ships Mondays and Thursdays only. Customer orders Monday morning. `get_handover_date()` returns Monday straight away, because Monday is already valid. We ask for 10 days. PostNL returns all 10. Customer picks Wednesday.

But the merchant's next handover is Thursday. Wednesday was never possible. Nothing catches it.

The V4 request has no field for excluded days, so this has to happen in `map_response()` instead: drop any date you can't reach from a valid drop-off day.

This is the biggest behaviour difference I found, and you can't see it from the diff.

src/Rest_API/V4/Timeframe/Service.php
```diff
@@ -0,0 +1,482 @@
...
+	 *
+	 * @param \Postnl\Sdk\ResponseData\V4\TimeFrame $timeframe SDK timeframe entry.
+	 *
+	 * @return string
+	 */
+	protected function map_option( $timeframe ): string {
+		$service = is_string( $timeframe->service ) ? strtolower( $timeframe->service ) : '';
+		$slot    = $timeframe->timeFrame;
+
+		if ( DeliveryWindowService::Evening->value === $service || ( null !== $slot && $slot->isEvening() ) ) {
```
tommyshellberg, line 398: Evening has no `is_evening_enabled()` check. Morning, two lines down, does. Two bugs come out of that.

**Evening is off but customers still get an evening option.**

`TimeSlot::isEvening()` just checks `from >= 17:00` (SDK, `src/ResponseData/V4/TimeSlot.php:45`). So a normal daytime window of 17:00-21:00 gets labelled `Evening` even when the merchant switched evening delivery off. That's an extra fee tab legacy would never show.

**Morning is off but it gets picked by default.**

With morning off, the 08:00-12:00 window doesn't get dropped, it gets renamed to `Daytime`. Then two things happen:

- `Delivery_Day.php:139` draws one radio button per timeframe, so the customer sees two "Daytime" options at the same price on the same date.
- `Container.php:257-275` throws out the options that carry a fee, then takes `$options[0]`. `Daytime` has no fee, so the renamed morning window survives and it's first in the list. It becomes the default.

So a merchant who turned morning delivery off gets morning selected by default, free, for every customer. That time slot is what gets saved and ends up on the label.

The comment on the morning branch already says why it has its check: don't show a fee tab legacy wouldn't. Evening needs the same check. In both cases, drop the window instead of renaming it.

src/Rest_API/V4/Timeframe/Service.php
```diff
@@ -0,0 +1,482 @@
...
+			return array( 'DeliveryOptions' => array() );
+		}
+
+		try {
+			$request  = $this->build_request( $post_data );
+			$client   = $this->build_client();
+			$response = $client->timeframes()->forMultipleServices( $request );
+
+			return array( 'DeliveryOptions' => $this->map_response( $response->timeframes() ) );
+		} catch ( \Throwable $exception ) {
```
tommyshellberg, line 178: Should this log something? Right now nothing on the V4 path writes to the log at all.

Four places it could:

- This catch converts and rethrows without writing anything.
- `Cache_Adapter` on line 223 is built without its optional `$logger`, so `log_bypass()` never fires. That warning exists to tell you the keyPrefix doesn't match the allowlist and nothing is being cached, which otherwise looks identical to an empty cache.
- `Client_Factory::make_builder()` doesn't attach a logger, so the SDK itself stays quiet.
- No request or response gets recorded.

Legacy logs all of it. `Rest_API\Base::send_request()` lines 227-252 writes the URL, the request args, the response headers, the response body, and any `WP_Error`. If a merchant reports a problem you can see exactly what happened.

One more thing makes it worse. `Exception_Converter::convert()` throws away the SDK's message and replaces it with a safe one, keeping the original in `getPrevious()`. Replacing it is right, you don't want API internals in a checkout notice. But nothing reads `getPrevious()`, so the real error is gone. And one of the messages it uses is "Check the PostNL logs for details." The logs will be empty.

`Logger_Adapter` already does this job. PSR-3 to `WC_Logger`, tags everything `[postnl-v4]`, has its own tests. Nothing in `src/` uses it, on any branch I checked. If the wiring PR hooks it up, add a line to the description. Otherwise V4 ships with no way to debug it.

src/Rest_API/SDK/Client_Factory.php
```diff
@@ -114,6 +115,34 @@ public function build(
...
+	 * (a caching plugin carries its own cache backend), so a fresh client is
+	 * assembled each call. The shared auth/version/customer/retry configuration
+	 * still comes from make_builder(), so callers add only the plugins they need.
+	 *
+	 * @param string              $v4_key     PostNL V4 API key.
+	 * @param bool                $is_sandbox Whether to target the sandbox environment.
+	 * @param HttpPluginInterface ...$plugins Transport plugins to attach in order.
+	 * @return PostnlClientInterface
+	 */
+	public function build_with_plugins( string $v4_key, bool $is_sandbox, HttpPluginInterface ...$plugins ): PostnlClientInterface {
```
tommyshellberg: `$v4_key` needs `#[\SensitiveParameter]`. `build()` on line 85 and `make_builder()` on line 160 both have it, and the docblock on lines 73-79 says why: PHP hides it from stack traces.

This method throws, and `Exception_Converter` keeps the original exception as `previous`, so the full trace survives. Without the attribute the API key can end up in `debug.log` and in the fatal error email WordPress sends the admin.

Also needs an `@since`. Everything else in this class has one.

tests/php/unit/Rest_API/V4/Timeframe/ServiceTest.php
```diff
@@ -0,0 +1,614 @@
...
+		$service = $this->make_handover_service( '2026-07-14 22:00:00', $this->make_settings() );
+
+		$this->assertSame( '2026-07-14', $service->expose_handover_date() );
+	}
+
+	/**
+	 * @testdox A malformed cut-off setting falls back to the 23:00 default instead of failing
+	 */
+	public function test_handover_tolerates_malformed_cutoff(): void {
+		$settings = $this->make_shipping_settings( 'not-a-time', '1', array( 'mon', 'tue', 'wed', 'thu', 'fri', 'sat', 'sun' ) );
```
tommyshellberg: This test doesn't check what its name says.

Line 313 compares times as strings. `'22:00' > 'not-a-time'` is `false`, because `n` is higher than `2`. So it takes the same branch whether the regex validates or not. I deleted the regex on `Service.php:355` and every test still passed.

Use `'16:0'` instead. Still malformed, but it sorts below the clock, so `'22:00' > '16:0'` is `true` and the two paths actually differ. Then deleting the regex fails this test.

I ran 23 mutations on `Service.php` while I was here. Six changed nothing:

- Skipping `Exception_Converter::convert()` entirely (line 140)
- Deleting either half of the evening check (line 285)
- Deleting the cut-off regex (line 355)
- Changing the drop-off loop from `< 6` to `< 2` (line 325)
- Removing the `$ttl > 0` fallback (line 427)
- Removing the null check (line 247)

The first one matters most. There's no `expectException` anywhere in this file, so the catch block and `Exception_Converter` have no test. `Counting_Http_Client` always returns 200, so no test ever makes a request fail. `Exception_ConverterTest.php` has fixtures you can copy.

Two smaller things:

- Line 430 checks for `array('Daytime')`, which is what `map_option()` returns when nothing else matches. That assertion passes no matter what the code does. Use a morning or evening window.
- Nothing tests that a different address misses the cache. A `Cache_Adapter` that ignored keys completely would pass every test here. The key itself is fine, `CachingPlugin::buildCacheKey()` hashes method + URI + body and the address is in the body, but no test shows that.

tommyshellberg:
Rather than another review round, I fixed my own review findings directly on the branch — seven commits, one per finding (`6efd1a2..17135d4`):

- **API key** — `get_v4_api_key()` doesn't exist anywhere, so the service always got an empty key. This turned out worse than my review said: the SDK *throws* on an empty key, so every V4 lookup would have failed the checkout outright. The key and the delivery-day count are now constructor-injected (no defaults), so a missed wiring step fails loudly instead of quietly doing the wrong thing.
- **Settings typing** — `$settings` is now typed against the real `Settings` class and all the `method_exists()` guards are gone. The unit bootstrap got a minimal `WC_Settings_API` stub so test doubles extend the real class; a wrong method name now breaks tests instead of passing against a double that invents the method. (This also exposed and fixed a pre-existing session-cache test that had invented its own `POSTNL_SETTINGS_ID` value.)
- **Evening/morning windows** — both are gated on their settings now, and disabled windows are dropped instead of renamed to Daytime. The renaming was creating a duplicate fee-free option that became the preselected default even when the merchant had switched morning delivery off.
- **Drop-off days** — delivery dates are now filtered across the whole window, not just the first handover. A Mon/Thu-shipping merchant no longer shows a Wednesday delivery date nobody can fulfil. The filter sits in the response mapping, so cached responses are filtered identically.
- **Logging** — the service now requires a PSR-3 logger: failures log the original SDK error (previously lost in `getPrevious()` while the safe message told you to check empty logs), the cache adapter gets the logger so its bypass warning can fire, and `Client_Factory` can attach a logger to the SDK builder. Wiring passes `Logger_Adapter` at flag-flip time.
- **`#[\SensitiveParameter]`** added on `build_with_plugins()`, and the `@since` tags aligned to 6.0.0.
- **Tests** — all six surviving mutations from my review are killed (including the vacuous `'not-a-time'` cut-off fixture and the missing exception-path coverage), plus a new test proving a different address misses the cache.

470 unit tests green, PHPCS clean, CI passing. I also retargeted the base from the stale `postnl-v4-timeframes-base` scaffolding branch to `main` so this merges to the right place.

**Ready to merge** — I'll squash-merge it.

One note for the follow-up branches: `postnl-v4-pickup-locations` and `postnl-v4-checkout-aggregation` carry the same `get_v4_api_key()` call and need the same constructor-injection fix (or a rebase onto this branch) before their reviews.

PR #338, opened 2026-07-17
Fix letterbox orders showing Standard Shipment as delivery type

sal4sup:
### All Submissions:

ClickUp: https://app.clickup.com/t/868kdd9c0

On the order edit page, Letterboxparcel orders showed **"Standard Shipment"** as the *Delivery Type* in the Label & Tracking box, even though the order was correctly shipped as Letterbox (the shipping line item and the "Letterboxparcel 48 hours" checkbox were both right).

Root cause: `Order\Base::get_delivery_type()` only maps delivery-day/pickup types via `Mapping::delivery_type()` and has no letterbox awareness, so a letterbox order falls through to the `Daytime` → "Standard Shipment" default.

Fix: `get_delivery_type()` now first checks the same resolved shipping options that drive the meta-box checkboxes and returns the letterbox label (`Letterboxparcel (48h)` / `(24h)`). The summary line therefore always agrees with the ticked checkbox and the shipping line item. Pure display fix — product-code selection (2928/2948) and label generation are unchanged.

Added 3 regression tests to `LetterboxTypeTest` and a `5.9.9` changelog entry.

1. In PostNL settings (NL store), set *Default automatic letterboxparcel product* to **Letterboxparcel 48**.
2. Place an order for a letterbox-eligible product (e.g. the "Sunglasses" test product).
3. Open the order in the admin. In the **Label & Tracking** box, confirm *Delivery Type* now reads **Letterboxparcel (48h)** (previously "Standard Shipment"), and that it matches the ticked "Letterboxparcel 48 hours" checkbox and the shipping line item.
4. Repeat with the default set to **Letterboxparcel Standard (24 hours)** and confirm *Delivery Type* reads **Letterboxparcel (24h)**.
5. Confirm a normal (non-letterbox) order still shows its usual delivery type (Standard Shipment / Evening / Pickup, etc.).

### Other information:

### Changelog entry

Fix: Order edit page showed "Standard Shipment" as the delivery type for Letterboxparcel orders instead of the selected 24-hour or 48-hour variant.

### DIFF
--- a/README.md
+++ b/README.md
@@ -29,8 +29,6 @@ PostNL’s official extension for WooCommerce on WordPress. Manage your national
 
 ### 5.9.9
 * Fix: Order edit page showed "Standard Shipment" as the delivery type for Letterboxparcel orders instead of the selected 24-hour or 48-hour variant.
-* Fix: Prevented a critical error on the order edit page for orders whose saved delivery date was in an unexpected format.
-* Fix: Plugin no longer fails to load on PHP 7.4-8.1 due to a development-only API SDK dependency being included in the production build.
 
 ### 5.9.8
 * Add: Letterboxparcel 48 hours as a selectable option in the domestic default shipping option, the bulk Change Shipping Options action, and the Label & Tracking menu.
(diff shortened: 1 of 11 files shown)


dustinparker (approved):
This fixes the bug. Letterbox orders now show the resolved 24h or 48h variant on the order edit page instead of "Standard Shipment", and because the summary and the checkbox read from the same source, the two can no longer disagree. Non-letterbox orders still resolve their delivery type through the existing mapping.

PR #339, opened 2026-07-18
16. V4 Pickup Locations + caching

sal4sup:
### All Submissions:

> The unchecked checkout boxes are **not applicable**: this PR adds a service class only. It is **not wired into checkout** — `Container::get_checkout_data()` still calls the timeframe half alone; the pickup half is connected in **task 17 (Checkout aggregation)**. Nothing runs unless a site has a validated V4 key AND the flag is on (default off), so merchant behavior is unchanged.

ClickUp: https://app.clickup.com/t/868jqaqgh — **16. V4 Pickup Locations + caching**

Adds `Rest_API\V4\Pickup_Location\Service`, the SDK-backed twin of the task-15 timeframe service, implementing `Pickup_Location_Service_Interface`:

- Builds a `PickUpNearAddressRequest` from the checkout address (via `Address_Utils::set_post_data_address`, mirroring `Legacy\Checkout\Item_Info`), requesting `Retail` locations only (parcel lockers were never surfaced by the legacy checkout).
- The V4 API key, the location count and a PSR-3 logger are **required constructor parameters** (same convention as the task-15 timeframe service after its review round): wiring in task 17 passes the key from `Settings::get_api_key_new()`, the count from `Settings::get_number_pickup_points()` (clamped to the V4 range **[1, 10]**), and a `Logger_Adapter`. Forgetting any of them is a TypeError at wiring time, not a silently unauthenticated client or a dead logger.
- Routes the call through the SDK `CachingPlugin` + `Cache_Adapter` with only `/locations/` allowlisted (`keyPrefix: Cache_Adapter::PREFIX_LOCATIONS`, so the prefix and the adapter allowlist cannot drift), reusing `Client_Factory::build_with_plugins()`. The adapter gets the logger so its allowlist-bypass warning can fire.
- Sends `pickupDate` as the **arrival** date (handover from the cut-off/transit/drop-off-day walk, plus one network day), matching the SDK's documentation of the field. The customer-facing group `PickupDate` is the same day formatted `d-m-Y` like every other date the plugin shows; the request stays ISO. Both come from one memoised date so they cannot disagree.
- Maps the flat V4 `locationsCollection()` into a single legacy `PickupOptions` group (`PickupDate` + `Locations[]`) with `LocationCode`, `Name`, `Distance`, `Address{CompanyName, Street, HouseNr, Zipcode, City, Countrycode}`, and flattened `OpeningHours`.
- Converts SDK errors through `Exception_Converter` to the legacy error shape, and logs the original cause (country + postcode area only — no street-level PII) before rethrowing.

**Mapping note — the full reconciliation set deferred to task 17**

V4 near-address does not return everything the legacy `/shipment/v1/checkout` `PickupOptions` shape carried, and the legacy consumers still expect all of it. Per the `Pickup_Location_Service_Interface` contract these are reconciled where the two checkout halves are aggregated (**task 17**), not in this service. Recording the complete set here so task 17's reviewer has it in one place:

- **`PartnerID`** — not returned by V4. `Frontend\Dropoff_Points::get_content_data()` (`src/Frontend/Dropoff_Points.php:84`) skips any location without it. It is also half the option *identity* in both modes — `sanitize_title( partner_id . '-' . loc_code )` in `templates/checkout/postnl-dropoff-points.php:114-115` and the same composite in `client/checkout/postnl-dropoff-points/block.js:164,177,260` — so task 17 has to choose an identity scheme, not just backfill the field.
- **`PickupTime`** — not returned by V4. Same `:84` guard. It is also persisted (`dropoff_points_time`) and rendered in the admin order box, the confirmation email, and passed to the label request (`Legacy\Shipping\Item_Info:221`), so whatever task 17 supplies has to be a real value, not a placeholder.
- **`Option` on the group** — not returned by V4. `Dropoff_Points.php:77` reads it into `type`, which becomes both a CSS class and the translated label text in the blocks component (`block.js:277,316`), and is persisted as `dropoff_points_type`.
- **`Distance` of 0 or unknown** — this service maps a null SDK distance to `''`, and `empty()` at `Dropoff_Points.php:84` is true for both `''` and `'0'`, so a location with no distance *and* a genuine distance of zero are both silently dropped. The unit is correct as mapped: the SDK returns metres (`?int $distance`) and `Utils::maybe_convert_km()` expects metres — do not "fix" this by switching to the SDK's `distanceInKilometers()` helper.
- **The pickup-enabled gate** — legacy only asked PostNL for pickup locations when `is_pickup_points_enabled()` was on (`Legacy\Checkout\Client.php:112-114` omits `'Pickup'` from `Options`). Nothing on the V4 path checks it, and the aggregation in #340 as currently written calls `get_pickup_locations()` unconditionally. Note this is more than a wasted request: `Frontend\Dropoff_Points::is_enabled()` has no callers anywhere, so the classic-checkout tab is purely response-driven (`Dropoff_Points.php:45-48`) and a non-empty `PickupOptions` would render a Pickup tab on stores that switched pickup off. Blocks gates client-side (`postnl-container/block.js:159,682`) and would only pay the extra call.
- **`PickupDate` format** — handled in this PR: the group date is `d-m-Y` (what customers, the admin box and the label request see), the request field is ISO. Task 17 has nothing to convert.
- **`OpeningHours` has no consumer yet.** It is mapped and tested, and the interface documents it as optional, but nothing in `src/`, `templates/`, `client/`, or `assets/` reads it — legacy never surfaced it either. It is carried because V4 returns it and dropping data at the mapping boundary is harder to undo than ignoring it downstream; task 17 does not need to do anything with it.

Until that set is reconciled, the `Dropoff_Points.php:84` guard skips every V4 location and both checkout modes would render an empty pickup tab. That is not reachable today: no V4 pickup service is registered via `inject_v4_service()`, `has_v4_key()` requires a validated new key that does not exist yet, and every `Router` flag defaults off.

**Depends on:** #337 (task 15) — **merged to `main`**; this PR is retargeted to `main` and carries the merge.

This PR has no runtime/checkout surface yet (wired in task 17). Verification is via the unit suite:

1. Check out the branch and run the unit tests: `npm run test:php:unit` (wp-env), or `vendor/bin/phpunit -c phpunit-unit.xml.dist` directly (no WordPress needed).
2. Confirm `Rest_API\V4\Pickup_Location\ServiceTest` passes: request building (address mapping, `numberOfLocations` clamp to [1,10], `Retail` type, arrival-date walk incl. exact-cut-off-minute and memoisation cases), response mapping to the `PickupOptions` shape (incl. empty + missing-optional-field cases, `d-m-Y` group date vs ISO request date), the caching round-trip, and the failure-path logging.
3. Confirm CI is green (unit + integration).

### Other information:

### Changelog entry

Internal (no merchant-facing change — behind the V4 SDK flag, default off): add the V4 SDK pickup-location lookup service with WP-transient caching.


src/Rest_API/V4/Pickup_Location/Service.php
```diff
@@ -0,0 +1,472 @@
...
+		$dropoff_days = $this->get_dropoff_days();
+		if ( ! empty( $dropoff_days ) ) {
+			$attempts = 0;
+			while ( $attempts < 6 && ! in_array( self::WEEKDAYS[ (int) $handover->format( 'N' ) ], $dropoff_days, true ) ) {
+				$handover = $handover->modify( '+1 day' );
+				++$attempts;
+			}
+		}
+
+		$this->pickup_date = $handover->format( 'Y-m-d' );
```
tommyshellberg: This returns the handover date, but everything around it wants the arrival date.

The SDK field this feeds is documented as "the date on which the parcel needs to be delivered at the PickUpLocation" (`PickUpNearAddressRequest.php`). This method's own docblock says "the date the parcel is expected to reach the location". And the Transit Time setting is "The number of days it takes for the order to be delivered after the order has been placed" (`Settings.php:423`), default 1.

Walk through the default config: customer orders Monday 10:00, cut-off 23:00, transit time 1, all drop-off days enabled. No cut-off shift, no extra days, Monday is enabled — so this sends **Monday**. But Monday is the day the merchant hands the parcel to PostNL. It can't be at the shop before Tuesday. Once task 17 wires this up, the customer is shown a pickup date the parcel can't make, in every configuration.

The walk itself is right — it lands the *handover* on an enabled drop-off day. It's just missing the last leg. Under this method's own model (transit − 1 preparation days, then handover, then one day in the PostNL network):

```php
$this->pickup_date = $handover->modify( '+1 day' )->format( 'Y-m-d' );
```

Two follow-ups worth a thought: if the walked handover is Saturday, the +1 lands on Sunday — does the V4 API do something sensible with a Sunday pickupDate? And `Timeframe/Service.php:313-331` on the base branch has the identical walk feeding `startDate`, so whatever shape the fix takes should land in both — task 17 shows both halves to the same customer from one response.

src/Rest_API/V4/Pickup_Location/Service.php
```diff
@@ -0,0 +1,472 @@
...
+	/**
+	 * Read the V4 API key from settings, tolerating settings without the getter.
+	 *
+	 * The V4 key field ships in a separate PR, so Service_Factory only routes here
+	 * once it exists; the method_exists guard keeps this class safe in isolation.
+	 *
+	 * @return string
+	 */
+	private function get_v4_key(): string {
+		if ( ! method_exists( $this->settings, 'get_v4_api_key' ) ) {
```
tommyshellberg: Same as the #337 comment on `Timeframe/Service.php:459`: `get_v4_api_key()` doesn't exist anywhere — #309 ships `get_api_key_new()`. So this always returns `''`, the SDK client is built without a usable key, and `Cache_Adapter`'s `sha1( $v4_key )` store prefix collapses to the same constant for every store. The suite stays green because both settings stubs in `ServiceTest.php` define `get_v4_api_key()` (lines 57 and 306).

One fix across the stack covers #337, this PR, and #340 — either the rename, or (better) constructor-inject the key from `Service_Factory`, which already knows how to read it. No need for three review rounds on this one.

src/Rest_API/V4/Pickup_Location/Service.php
```diff
@@ -0,0 +1,472 @@
...
+		try {
+			$request  = $this->build_request( $post_data );
+			$client   = $this->build_client();
+			$response = $client->pickupLocations()->nearAddress( $request );
+
+			return array( 'PickupOptions' => $this->map_response( $response->locationsCollection() ) );
+		} catch ( \Throwable $exception ) {
+			// Exception_Converter returns a plugin-shaped \Exception; its message can
+			// carry raw API text (field errors, upstream messages) — escape on output.
+			$error = Exception_Converter::convert( $exception );
```
tommyshellberg, line 186: Same question as on #337: should the V4 path log? This catch converts and rethrows without writing anything, and the `Cache_Adapter` on line 213 is built without its optional logger, so its `log_bypass()` warning can never fire. Legacy `Base::send_request()` writes the URL, request and response to the log on every call when the merchant's logging toggle is on; on the V4 path that toggle silently does nothing. `SDK/Logger_Adapter.php` is finished and tested — it's just never handed to anything.

src/Rest_API/V4/Pickup_Location/Service.php
```diff
@@ -0,0 +1,472 @@
...
+	 * @return \Postnl\Sdk\Client\PostnlClientInterface
+	 */
+	protected function build_client() {
+		$v4_key = $this->get_v4_key();
+
+		$caching_plugin = CachingPlugin::create(
+			cache: new Cache_Adapter( $v4_key ),
+			ttl: $this->cache_ttl(),
+			allowedEndpoints: array( '/locations/' ),
+			keyPrefix: 'locations'
```
tommyshellberg: `Cache_Adapter::PREFIX_LOCATIONS` exists for exactly this string (`Cache_Adapter.php:66`) — using it means the prefix and the adapter's allowlist can't drift apart. Minor: I mutated this to `'foo'` and `test_second_identical_call_hits_cache` went red, so drift at least fails CI.

src/Rest_API/V4/Pickup_Location/Service.php
```diff
@@ -0,0 +1,472 @@
...
+			$locations[] = $this->map_location( $location );
+		}
+
+		if ( empty( $locations ) ) {
+			return array();
+		}
+
+		return array(
+			array(
+				'PickupDate' => $this->get_pickup_date(),
```
tommyshellberg: Two consumers get this one string and they want different formats. The request side (`pickupDate`, line 167) needs ISO `Y-m-d` — right as is. But this `PickupDate` ends up in front of people: it's echoed raw in the classic checkout template (`templates/checkout/postnl-dropoff-points.php:137`), the admin order box, and the confirmation email — and every other date in the plugin is `d-m-Y`. The V4 timeframe service deliberately kept `dd-mm-yyyy` for delivery dates for exactly this reason. Format the group date as `d-m-Y` here so task 17 has nothing to convert and customers don't suddenly see ISO dates.

src/Rest_API/V4/Pickup_Location/Service.php
```diff
@@ -0,0 +1,472 @@
...
+ * near-address endpoint, and maps the response back into the PickupOptions shape
+ * Frontend\Container consumes (see Pickup_Location_Service_Interface), so callers
+ * cannot tell V4 from the legacy /shipment/v1/checkout path.
+ *
+ * V4 near-address returns a flat list of locations rather than the legacy
+ * per-date PickupOptions groups, so the whole list is emitted as a single group
+ * keyed on the computed pickup date.
+ *
+ * V4 provides no PartnerID or PickupTime for a location; those legacy-only fields
+ * are reconciled where the two checkout halves are aggregated (task 17), not here.
```
tommyshellberg, line 45: The deferred list is longer than `PartnerID` and `PickupTime`. From reading `Dropoff_Points::get_content_data()` and the blocks component, task 17 also has to reconcile:

- **`Option` on the group** — `Dropoff_Points.php:77` reads it into `type`, which becomes a CSS class and label text in the blocks component. V4 groups don't carry it.
- **`Distance`** — `Dropoff_Points.php:84` skips a location when `Distance` is empty, and this service maps a null SDK distance to `''` (line 270). A location with no distance is silently dropped — and so is a genuine distance of 0.
- **The pickup-enabled gate** — legacy only asked for pickup locations when `is_pickup_points_enabled()` (`Legacy/Checkout/Client.php:112-114`). Nothing on the V4 path checks it, so the aggregation needs that gate or merchants who switched pickup off still pay a locations API call on every checkout.

Fine to defer all of it — but please add these to the PR description alongside PartnerID/PickupTime so task 17's reviewer knows the full set. As it stands, the `Dropoff_Points.php:84` guard skips every V4 location (no PartnerID, no PickupTime), so both checkout modes render an empty pickup tab until the whole set is reconciled.

src/Rest_API/V4/Pickup_Location/Service.php
```diff
@@ -0,0 +1,472 @@
...
+	private $pickup_date = null;
+
+	/**
+	 * Service constructor.
+	 *
+	 * @param Client_Factory $client_factory      SDK client factory.
+	 * @param object         $settings            Plugin settings instance.
+	 * @param int            $number_of_locations Locations to request (default 3, capped at the V4 max of 10).
+	 */
+	public function __construct( Client_Factory $client_factory, object $settings, int $number_of_locations = self::DEFAULT_LOCATIONS ) {
```
tommyshellberg: Harmless today — `Settings::get_number_pickup_points()` is hardcoded to return 3, same as this default. But the task-17 wiring should still pass `$settings->get_number_pickup_points()` explicitly: that getter has a commented-out country option ("Temporarily hardcoded"), and if it ever comes back, legacy follows the merchant's setting while V4 silently stays at 3. Same shape as the `number_delivery_days` comment on #337, just with no visible consequence yet.

tests/php/unit/Rest_API/V4/Pickup_Location/ServiceTest.php
```diff
@@ -0,0 +1,639 @@
...
+		$service  = $this->make_pickup_date_service( '2026-07-17 17:00:00', $settings );
+
+		$this->assertSame( '2026-07-20', $service->expose_pickup_date() );
+	}
+
+	/**
+	 * @testdox A malformed cut-off setting falls back to the 23:00 default instead of failing
+	 */
+	public function test_pickup_date_tolerates_malformed_cutoff(): void {
+		$settings = $this->make_shipping_settings( 'not-a-time', '1', array( 'mon', 'tue', 'wed', 'thu', 'fri', 'sat', 'sun' ) );
```
tommyshellberg: Same fixture as #337: this test can't tell the validated path from the unvalidated one. `Service.php:339` compares times as strings, and `'22:00' > 'not-a-time'` is false either way — I deleted the regex on `Service.php:383` and this test stayed green. Use `'16:0'`: still malformed, but it sorts below the clock, so the two paths produce different dates and the test starts proving something.

tests/php/unit/Rest_API/V4/Pickup_Location/ServiceTest.php
```diff
@@ -0,0 +1,639 @@
...
+		$service = new Pickup_Date_Service( new Client_Factory( $settings ), $settings );
+		$service->set_now( new \DateTimeImmutable( $now ) );
+
+		return $service;
+	}
+
+	/**
+	 * @testdox An order after the cut-off time is picked up the next day
+	 */
+	public function test_pickup_date_after_cutoff_shifts_a_day(): void {
```
tommyshellberg, line 373: Two small gaps from mutation testing, both optional. Flipping the cut-off comparison on `Service.php:339` from `>` to `>=` survives every test — no case is pinned at exactly the cut-off minute. And removing the pickup-date memoisation survives too: nothing proves the request and the mapped group carry the same date, which is the whole reason the memoisation exists. A test through `get_pickup_locations()` whose `now()` returns 22:59 on the first call and 23:01 on the second (cut-off 23:00) would fail without the memo and pass with it.

tommyshellberg (changes requested):
One real behaviour bug in the new code (the pickup date is the handover date, not the day the parcel reaches the shop), plus the two items already raised on #337 that recur here verbatim (`get_v4_api_key`)

**Hard gates — all pass:**

| Gate | Result |
|---|---|
| SDK in `require-dev`, pinned `dev-main#5a064dd`, `php >= 7.4` floor | ✅ |
| No leakage into legacy paths | ✅ — the diff is two new files; `Order\Base`, `Service_Factory`, `Client_Factory` untouched |
| Order meta shape / public filter signatures | ✅ unchanged (only reads the existing `postnl_v4_cache_ttl` filter) |
| Routing off by default | ✅ — nothing in production even calls `pickup_location_service()` yet; `Frontend\Container` still only calls the timeframe service |

**Checked and fine, so you don't have to worry about it:**
- The response cache can't serve one customer's locations to another — established on #337: the cache key hashes method + URI + body, and the address is in the POST body.
- `get_cut_off_time()`, `get_transit_time()`, `get_dropoff_days()`, `get_customer_code()`/`get_customer_num()` all exist on today's `Settings`, so those `method_exists()` guards resolve in production. Only `get_v4_api_key` doesn't (inline comment).
- Unit suite green (442 tests), PHPCS clean on the service. I ran 10 mutations against the suite: 6 caught, including the cache keyPrefix and allowlist (changing either makes the cache test fail — good), the mapping keys, and the clamp. The 4 survivors are covered in inline comments.
- `map_opening_hours()` is mapped and tested, but nothing in the plugin reads `OpeningHours` — no consumer in `src/`, `templates/`, `client/`, or `assets/` renders it. Fine to keep; just so you know it's currently write-only.

**Process notes:**
- This branch is behind `postnl-v4-timeframes`, which has merged main twice since (bringing #335/#336). After the #337 fixes land, merge down and confirm `Service_Factory::barcode_from_label()` and its call site in `Order/Base.php` arrive together — they live in different files and a partial merge resolution fatals on label save.
- Before squash-merging, the base must be retargeted to `main`. If #337 merges first, double-check GitHub's automatic retarget actually points here at `main` and not at a scaffolding branch.

tommyshellberg:
Review fixes are in, same treatment #337 got. `main` (with the merged task 15) is merged into this branch — the five conflicts were all stale copies from the old scaffolding base with no edits on this side, so they resolve to main's version. Base is retargeted to `main`.

The fix commits, one per finding:

- `df6b3f1` — the V4 key, the location count and a PSR-3 logger are now required constructor parameters. `get_v4_api_key()` never existed, so the old guard left the key permanently `''` — and `Auth::apiKey('')` throws, so every lookup would have failed checkout outright. The test doubles that defined `get_v4_api_key()` themselves are gone; they extend the real `Settings` now. Failures also get logged before the rethrow (country + postcode area only, no street-level data), and `Cache_Adapter` gets the logger so its bypass warning can fire.
- `7a79bcb` — `pickupDate` is now the arrival date: the walk still lands the handover on an enabled drop-off day, then one network day is added. Same model as the timeframe service's `get_handover_date()`. One correction to my own review comment: the "land the fix in both halves" follow-up is wrong post-merge — timeframes sends `handoverDate`, which genuinely is a handover date, and `is_reachable_delivery_date()` on main already accounts for the transit leg. Adding a day there would shift every delivery date late. The fix belongs here only.
- `89ef81b` — the group `PickupDate` customers see is `d-m-Y` like every other date in the plugin; the request stays ISO. Both formats come from the one memoised date so they can't disagree.
- `41c3998` — `keyPrefix` uses `Cache_Adapter::PREFIX_LOCATIONS` instead of repeating the literal.
- `bcf6854` — the `'not-a-time'` fixture is `'16:0'` now, so the malformed-cut-off test actually fails when the validation is deleted (it didn't before). New tests pin the exact-cut-off-minute boundary and prove the request date and the group date agree when the clock crosses the cut-off mid-lookup.

Every fix was mutation-checked: reverting the arrival fix fails 8 tests, deleting the cut-off validation fails the malformed-cut-off test, dropping the log call fails the logging test, forcing an empty key errors 3 tests with the exact production symptom, and the cut-off guard was mutated both ways (`>` → `>=` and never-fires) with different tests catching each direction. Suite is 489 unit tests, green; PHPCS clean.

Two items deserve #340's attention: `PartnerID` is half the option identity in both checkout modes, not just a guard field; and `Frontend\Dropoff_Points::is_enabled()` has no callers, so on the V4 path a merchant who switched pickup off would still get a Pickup tab in classic checkout. #340 currently calls `get_pickup_locations()` unconditionally.

tommyshellberg (approved):

tommyshellberg (approved):

PR #340, opened 2026-07-19
17. Checkout aggregation (compose delivery days + pickup points from two services)

sal4sup:
### All Submissions:

Task 17 of the V4 SDK migration. V1 returned delivery days and pickup points in a single `/shipment/v1/checkout` response; V4 splits them into two endpoints. This reconciles that at the consumer.

`Frontend\Container::get_checkout_data()` now composes the response from both flows via a new `aggregate_delivery_options()` method:

- It fetches delivery days from `Service_Factory::timeframe_service()`.
- The Legacy transport resolves both flows to **one shared** `Checkout_Service` whose single response already carries `DeliveryOptions` and `PickupOptions`, so when the pickup service is the same instance the second call is skipped — legacy keeps making exactly one API call, no behavior change.
- The V4 transport hands back a **distinct** `Pickup_Location` service, so its `get_pickup_locations()` result is fetched and its `PickupOptions` merged into the delivery-day response.
- The assembled array is identical either way, so `Frontend\Delivery_Day` / `Frontend\Dropoff_Points` and the blocks components render the same regardless of transport.

`Checkout_Blocks\Extend_Block_Core` already funnels through `Container::get_checkout_data()`, so the blocks path is covered without duplicated logic.

Notes:
- Error semantics match the legacy single call: either service throwing aborts the whole lookup, so a failure never renders delivery days without pickup points or vice versa.
- V4 lookups are issued sequentially — the plugin/SDK have no async transport, so true parallel execution is deferred; repeat pageloads for the same address hit the SDK response cache, so the second call is cheap.
- Everything stays behind the per-flow feature flags (default off); with a V4 key + flags off, the shared legacy instance is used and nothing changes for merchants.

Depends on #337 (Task 15 — timeframes) and #339 (Task 16 — pickup locations); based on `postnl-v4-pickup-locations`.

Closes https://app.clickup.com/t/868jqaqgn

1. **Flags off (default):** With any configuration, load classic and blocks checkout with a NL address — delivery days and pickup points render exactly as before (one checkout API call).
2. **V4 flags on (staging, with a validated New API Key):** enable `postnl_sdk_enable_timeframe` and `postnl_sdk_enable_pickup_location`, then load classic and blocks checkout with a NL address — delivery days and pickup points both render identically to the legacy path.
3. Run the unit suite: `npm run test:php:unit` — `ContainerCheckoutAggregationTest` covers the single-call legacy path, the two-call V4 merge, and the empty-pickup fallback.

### Other information:

### Changelog entry

Not applicable — internal V4 migration step, fully behind per-flow feature flags (default off), no merchant-facing change (consistent with the sibling V4 subtask PRs).


src/Frontend/Container.php
```diff
@@ -209,6 +215,52 @@ public function get_checkout_data( $post_data ) {
...
+	protected static function aggregate_delivery_options( Timeframe_Service_Interface $timeframe_service, Pickup_Location_Service_Interface $pickup_service, $post_data ) {
+		$response = $timeframe_service->get_delivery_options( $post_data );
+
+		// Legacy shares one instance across both flows; its single response already
+		// holds both halves, so a second call would only repeat the same request.
+		if ( $pickup_service === $timeframe_service ) {
+			return $response;
+		}
+
+		$pickup                    = $pickup_service->get_pickup_locations( $post_data );
```
tommyshellberg, line 269: `get_pickup_locations()` runs even for merchants who turned pickup points off.

Legacy only asks PostNL for pickup points when the setting is on: `Base_Info.php:95` copies `Settings::is_pickup_points_enabled()` into the request data, and `Legacy/Checkout/Client.php:112` only adds `'Pickup'` to the request options when it is set. A pickup-disabled shop gets a response with no `PickupOptions`, so the pickup tab stays hidden.

On the V4 path this line always queries the pickup service whenever it is a separate instance. For a shop with pickup points disabled that means:

- the pickup tab comes back at checkout,
- every address lookup makes an API call the merchant switched off,
- and if that call fails, the `throw` takes the delivery days down with it — a total checkout-options outage caused by a feature the merchant isn't using.

Suggested fix — pass the setting in from `get_checkout_data()` (which has `$this->settings`) and skip the call entirely, leaving the key absent exactly as legacy does:

```php
if ( $pickup_service === $timeframe_service || ! $pickup_enabled ) {
    return $response;
}
```

(Delivery days don't need a matching gate — `Base_Info.php:96` hardcodes `delivery_days_enabled` to `true`, so always fetching those is exact parity.)

tommyshellberg (commented):
One problem to fix before merge — the pickup lookup ignores the merchant's pickup-points setting (inline comment on line 258). Everything else I checked came back fine.

**What I verified**

- **Legacy checkout still makes exactly one API call.** Both `Service_Factory::timeframe_service()` and `pickup_location_service()` fall through to `legacy_checkout_service()`, which memoises one shared `Legacy\Checkout_Service` under the `'checkout'` key — so the `===` identity check in the aggregator genuinely holds and legacy returns before the second call. The call-count test pins this.
- **The real V4 service matches the test doubles' contract.** `V4/Pickup_Location/Service::get_pickup_locations()` returns `array( 'PickupOptions' => ... )` (Service.php:182), which is the key the isset merge reads. I checked this against the real class because hand-written doubles have hidden a contract mismatch in this queue before.
- **Mixed flag configurations work.** Timeframe-V4 + pickup-legacy gives two distinct instances, and the legacy `get_pickup_locations()` fires the full checkout call whose response carries `PickupOptions`. The reverse (timeframe-legacy + pickup-V4) overwrites the legacy pickup list with V4's, which is the intended behavior.
- **Delivery days need no matching gate.** `Base_Info.php:96` hardcodes `delivery_days_enabled` to `true`, so the timeframe half always fetching is exact legacy parity.
- **Mutation testing: 7 of 7 killed.** Deleting the `===` guard fails the legacy call-count test; flipping it to `!==` fails all three; dropping the isset fallback, deleting the merge, returning the pickup response, ignoring the pickup result, and inverting the ternary each fail the composition or missing-key test. The guard was mutated both ways (never-fires and always-fires) and those fail *different* tests, so none of the assertions are decorative.
- **PHPCS**: the 10 errors / 16 warnings in `Container.php` all pre-exist on main (identical count on main's copy); the new lines and the test file are clean.
- **Hard gates**: `composer.json` untouched (SDK stays in require-dev), `Service_Factory`/routing untouched, no order-meta or filter-signature changes. CI green on both suites.

src/Frontend/Container.php
tommyshellberg: Fixed in ebc2596. `aggregate_delivery_options()` now takes a `bool $pickup_enabled` (passed from `Settings::is_pickup_points_enabled()` at the call site) and the guard skips the lookup outright when it's off, leaving `PickupOptions` absent exactly as legacy does.

Verified: new test `test_disabled_pickup_points_skip_the_lookup_entirely` asserts zero pickup calls **and** no `PickupOptions` key (its double returns a non-empty list, so a leak trips both assertions). Written red-first against the un-gated code, then mutated both ways — a gate that never fires fails the new test, a gate that always fires fails the two composition tests, disjoint sets. Full unit suite 493 green; PHPCS unchanged from the pre-existing baseline.

PR #342, opened 2026-07-21
Fix Letterbox 48 reverting to Standard Shipment on legacy-storage bulk label creation

sal4sup:
### All Submissions:

Fixes **Letterbox 48 orders reverting to Standard Shipment** when a label is generated via the **bulk action** on the classic **WordPress posts (non-HPOS) orders list**. HPOS stores were unaffected, which is why the issue reproduced on one environment and not the other.

Root cause: `Order\Base::save_meta_value()` decided which fields to persist using the same field list that the bulk **modal** trims for display (`Bulk::additional_meta_box`). On the legacy list screen that trim removes every `show_in_bulk => false` field — including **Letterbox 48** (and ID Check). The injected Letterbox 48 selection was therefore never saved, so the label engine fell back to the Standard product and both the label and the order display flipped to Standard.

Fix: give `meta_box_fields()` a `$context` (`'display'` vs `'save'`) that is passed through the `postnl_order_meta_box_fields` filter. `Bulk::additional_meta_box()` now trims only for `'display'` and returns the full field set for `'save'`, so persistence can no longer drop a selected option. This keeps the display-trim behaviour of the bulk modal intact while making the save path storage-mode agnostic. Also fixes the same latent drop for ID Check on legacy bulk.

Relates to ClickUp task: https://app.clickup.com/t/4550855/868kdd9c0

Closes # .

1. On a store using **WordPress posts storage (legacy / HPOS off)**, set PostNL → *Default automatic letterboxparcel product* to **Letterboxparcel 48 hours**, and have a Letterbox-eligible product.
2. Place an order for that product (NL). Confirm it shows **Letterbox 48** in the orders list and in the order's *Label & Tracking* box.
3. From the classic Orders list, tick the order → bulk action **PostNL Create Label** → **Apply**.
4. Open the order: **Delivery Type stays "Letterboxparcel (48h)"** and the generated label is the Letterbox 48 product (2948) — **not** Standard Shipment. (Before this fix it reverted to Standard.)
5. Turn **HPOS on** and repeat 2–4 to confirm no regression on the new storage.
6. Confirm **24h Letterbox** and the **single-order "Create Label"** button are unchanged.

### Other information:

### Changelog entry

Fixed Letterbox 48 shipments reverting to Standard Shipment when a label was generated via the bulk action on the classic (non-HPOS) orders list.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -319,7 +319,37 @@ JS tests use Jest with `@testing-library/react`. Mocks in `tests/js/__mocks__/` 
 - `@wordpress/components`
 - `@woocommerce/settings`
 
-PHP has no unit test suite — only PHPCS linting. All PHP testing is manual.
+PHP has two PHPUnit suites, both run in CI (`.github/workflows/php-tests.yml`):
+
+- **Unit** (`tests/php/unit`, `phpunit-unit.xml.dist`) is pure PHP with no WordPress bootstrap, so it runs without wp-env. Run it with `npm run test:php:unit`.
+- **Integration** (`tests/php/integration`, `phpunit-integration.xml.dist`) needs a real WordPress and runs inside the wp-env `tests-cli` container. Run it with `npm run test:php:integration`.
+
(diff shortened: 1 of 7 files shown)


dustinparker (approved):
This fixes the bug. Letterboxparcel 48-hour orders now hold their delivery type through the bulk label action on the classic orders list, because the save path no longer inherits the bulk modal's display trim, and the same silent drop is closed for ID Check.

PR #343, opened 2026-07-21
18. V4 shipping label — domestic parcel core

sal4sup:
### All Submissions:

* [ ] Have you tested both blocks/non-blocks cart/checkout? _(N/A — this is admin label generation, not checkout. See note below.)_
* [ ] Have you tested the cart and checkout as both a logged-in user and a guest? _(N/A)_
* [ ] Have you successfully placed an order as both a logged-in user and a guest? _(N/A)_
* [ ] Did you have the query monitor plugin active during all testing? _(N/A — see testing note)_

ClickUp task 18: https://app.clickup.com/t/868jqaqgv (part of the V4 SDK migration, parent https://app.clickup.com/t/868jpv52c).

Adds the first V4 SDK service — outbound shipping labels — for the **happy-path single domestic NL parcel only**, behind the existing two gates (a validated V4 key **and** `postnl_sdk_enable_label`, both default off). Merging changes nothing for merchants.

- `src/Rest_API/V4/Label/Request_Builder.php` — pure translator from parsed order fields to a `ShipmentDeliveryRequest` DTO for `/shipment/delivery/v4/labelconfirm` (`shipmentDelivery()->labelConfirm()`). Maps sender/receiver `ShipmentParty`, `LabelSettings`, `ShipmentType::Parcel`, and one collo. Maps the legacy combined printer strings (`GraphicFile|PDF`, `Zebra|Generic ZPL II 600 dpi`) to discrete output-type + resolution. **No `CollectionLocation`, no `MessageID`, no product code** (resolved server-side from type + services).
- `src/Rest_API/V4/Label/Response_Mapper.php` — pure reader that captures the barcode and label document(s) from the `labelconfirm` response.
- `src/Rest_API/V4/Label/Service.php` — the WooCommerce-bound service (implements `Label_Service_Interface`). Uses `V4_Mapper` to resolve `ShipmentType` + `Services`; writes labels into the existing `_postnl_order_metadata['labels']` shape (`type`, `barcode`, `created_at`, `filepath`, `merged_files`) **tagged `api_version = v4`**, reusing `Order\Base::maybe_merge_labels()` so A4/A6 handling is unchanged. Anything outside the happy path (services, delivery-day/pickup, returns, multi-collo, international, non-3085 product) **falls back to the untouched legacy pipeline** — those flows migrate in tasks 19–21.
- `src/Rest_API/Service_Factory.php` — `label_service()` now lazily builds the V4 service when the label flow routes to V4 (test doubles injected via `inject_v4_service()` still win). The per-combination `V4_Mapper` gate lives in the service, where the resolved product exists.

Automated (runnable now):

1. `npm run test:php:unit` — all unit tests pass, including the new `Request_BuilderTest` (asserts the serialized DTO: sender/receiver/label/one collo, weight in grams, **no CollectionLocation/MessageID/product code**, empty barcode omitted so labelconfirm auto-issues, enum fallbacks, printer-string mapping) and `Response_MapperTest` (barcode capture, label extraction/decode, V4-tagged record shape).
2. `composer check-php` — no new PHPCS violations in the added files.
3. Regression: with the flag off (default), generate a label on any order — it goes through the legacy path exactly as before.

Staging (end-to-end V4 call): the live path needs the separate "New API Key" settings PR (adds `get_api_key_new()` / `is_api_key_new_validated()`, not on this branch yet) plus a V4 sandbox key, then `add_filter( 'postnl_sdk_enable_label', '__return_true' )`. Generate a label on a plain domestic NL order (single collo, standard delivery, no options/returns/pickup) and confirm a PDF is produced and the barcode is stored — then compare against a V1 baseline order (staging parity check).

### Other information:

**Stacked PR:** based on `postnl-v4-sdk-exception-converter` (task 13, its dependency). Merge that first; GitHub will retarget this to `main` afterwards.

**Decision to review (barcode):** the task note says labelconfirm auto-issues the barcode. Because the barcode flow itself isn't migrated yet (task 14), `Order\Base::save_meta_value()` still generates a barcode via the legacy Barcode API and persists it in `barcodes[]`. To keep `barcodes[]` and the label's `barcode` in sync, this PR **supplies that pre-issued barcode on the request** (the SDK's `ShippingItem->barcode`, mirroring the V1 barcode-then-confirm pattern) and captures the response barcode for the label record. When task 14 lands, the pre-issue step can be dropped and the auto-issued barcode used directly. Happy to switch to pure auto-issue here if preferred.

### Changelog entry

No changelog entry: internal, feature-flagged, and no merchant-facing behavior change (per the V4 migration epic ground rules — flags default off).

---

## Review fix round (2026-08-06)

Five commits landed addressing pullrequestreview-4872527804: `1f388ed` (bulk-crash fix: self-built V4 service no longer flips `barcode_from_label()`), `9997824` (constructor-injected key + required PSR-3 logger, failure logging with PII-trimmed context), `7a377fd` (mutation-survivor test fixes), `2390ada` (`@since` 6.0.0), `2b33c58` (`zpl` merge case). Unit suite 548 green.

**Staleness note on the sections above:** the "Stacked PR" paragraph and the "task 14" barcode decision predate the main merge — the base is now `main` and task 14 already landed there as #336 (`barcode_from_label()` + harvest path). The pre-issued-barcode flow described above remains the behavior of this PR: with nothing injected into `Service_Factory::$v4_services`, `barcode_from_label()` stays false and the legacy prefetch keeps `barcodes[]` and the label's barcode in sync. The wiring task switches to auto-issue by explicitly injecting the label service.

**Deferred: `postnl_shipment_addresses` does not fire on the V4 path.**
The filter has been a documented third-party extension point since 5.7.0 and is applied at `src/Rest_API/Legacy/Shipping/Client.php:292`. The V4 label path builds its request from `$item_info->shipper` / `->receiver` directly (`V4/Label/Service::extract_fields()`, `V4/Label/Request_Builder::build()`) and never applies it, so a third-party address rewrite would not reach a V4-routed label. The other label-path filter, `postnl_order_weight`, is inherited for free because the V4 service reuses the shared `Shipping\Item_Info`.

This is deferred to #352 (the filter carry-forward audit, task 25), which adds `Request_Builder::apply_filtered_receiver()` plus a `filter_shipment_addresses()` step that reuses the legacy address builder verbatim so the filter keeps its `( array $addresses, Shipping\Client $client )` signature. It is not implemented here because #352 already carries the implementation, an integration test and the carry-forward matrix, and duplicating it on this branch would put two variants of the same code in the same hunks that #352 has to rebase through.

Nothing ships to merchants in the meantime: V4 label routing needs both a validated V4 key and `postnl_sdk_enable_label`, and both default off. The gap is tracked as S15 in `docs/postnl-v4-migration/flip-checklist.md` and is a hard pre-flip gate ("`postnl_shipment_addresses` fires and an address rewrite reaches the PostNL request").

**Decided limitation: the "Shipping confirmation e-mail address" setting is not sent on the V4 path.**

On V1 the plugin appends a second contact to the shipment when `Settings::get_shipping_confirmation_email()` is non-empty (`src/Rest_API/Legacy/Shipping/Client.php:106-112`):

```
'Contacts' => [
    [ 'ContactType' => '01', 'Email' => <customer>, 'SMSNr' => <customer phone> ],
    [ 'ContactType' => '02', 'Email' => <merchant confirmation address> ],
]
```

The V4 SDK has no way to express the second entry. Verified exhaustively against `vendor/postnl/api-client-sdk`:

- `RequestData/V4/ShipmentParty` carries a single nullable `Contact` — no contacts array, no additional-contact field.
- `RequestData/V4/Contact` has one `email`.
- `ShipmentDeliveryRequest`'s payload is `sender`, `receiver`, `labelSettings`, `returnOptions`, `shipmentType`, `handoverDate`, `deliveryLocation`, `services`, `internationalShipmentData`, `items`, `itemCount` — no contacts block.
- `Enums/PayloadKey` has 167 cases; `email` is the only email-shaped one, and it belongs to `Contact`. No `contactType`, `contacts`, `notification` or `sms` key exists anywhere in the SDK.
- There is no escape hatch: `ShipmentDeliveryRequest::toArray()` emits an explicit key list and the nested DTOs normalize via reflected `#[Field]` attributes only, so an unknown field cannot be injected without patching the vendored SDK.

Consequence, once the label flag is flipped: a merchant who configured the address stops receiving PostNL shipping confirmations for V4-routed orders. Orders that fall back to legacy (services, international, multi-collo, pickup, delivery-day, return) keep sending it, so the loss is selective rather than total. Nothing errors and nothing is logged.

Nothing changes for merchants on merge — both gates default off. Tracked as S14 in `docs/postnl-v4-migration/flip-checklist.md`; an answer from PostNL is a pre-flip gate. If the answer is "no V4 equivalent", the setting must be marked V1-only in its description before the flip so merchants are not left believing it still works.


sal4sup:
**Self-review applied (dual-engine: WooCommerce standards + correctness/security).**

Fixed before requesting review:

- 🔴 **Silent feature drop:** `map_combination()` hardcoded `options => []`, so the mapper always resolved the base `3085` row — an insured (or delivery-code + insured) parcel keeps code `3085` and would have routed to V4 with insurance dropped. Now the real selected options are passed to `V4_Mapper`, plus an explicit `shipping_product['options']` guard, so those combinations resolve to a services row / unknown combination and fall back to legacy. Covered by new tests.
- 🟡 **Test gap:** extracted the eligibility + mapping decision into a pure `Eligibility` class and added `EligibilityTest` (base parcel eligible; multi-collo, delivery-day, pickup, return, product-options, evening, non-NL, no-v4-equivalent, mapped-services, pickup-location all fall back; insured and delivery-code+insured 3085 stay legacy).
- 🟢 **File write:** `store_labels()` now verifies the label file exists on disk before recording it, so a failed write never persists a meta entry pointing at a missing file.
- Standards: `Request_Builder` pure methods made `static`; SDK interfaces imported instead of referenced by FQN in docblocks.

Informational (no change): the barcode stays aligned by the pre-issue-and-echo design (see PR description); sender contact fields are intentionally omitted (the SDK injects the customer credentials and labelconfirm doesn't require sender contact).

Full unit suite green (414 tests); PHPCS clean on all changed source files.

src/Rest_API/Service_Factory.php
```diff
@@ -182,7 +183,13 @@ public function pickup_location_service(): Pickup_Location_Service_Interface {
 	 * @return Label_Service_Interface
 	 */
 	public function label_service(): Label_Service_Interface {
-		if ( $this->should_use_v4( 'label' ) && isset( $this->v4_services['label'] ) ) {
+		if ( $this->should_use_v4( 'label' ) ) {
+			// A test double injected via inject_v4_service() wins; otherwise build the real V4 service.
+			// The per-combination V4_Mapper gate lives inside the service, which falls back to the
+			// legacy pipeline for anything outside the happy-path domestic parcel.
+			if ( ! isset( $this->v4_services['label'] ) ) {
+				$this->v4_services['label'] = new V4_Label_Service();
```
tommyshellberg: This line breaks bulk label generation once the flag is on: the first order of a bulk run works, every later order fails with "Barcode is empty!".

Here's the chain. `barcode_from_label()` (line 152) returns `should_use_v4('label') && isset( $this->v4_services['label'] )`, and `Order\Base::save_meta_value()` calls it **before** any label call (`Order/Base.php:629`) to decide whether to prefetch a barcode. This line makes `label_service()` write into that same array. The factory is memoised per `Order\Base` instance (`Order/Base.php:99-103`), and one `Order\Bulk` instance handles a whole bulk loop. So:

- Order #1: `barcode_from_label()` is false (array still empty) → legacy barcode prefetch → `create_label()` → this line caches the V4 service. Works.
- Order #2 onward: `barcode_from_label()` is now true → no prefetch, no `main_barcode` in `$post_data` → `Service::create()`'s first line, `new Shipping\Item_Info( $post_data )`, throws "Barcode is empty!" (`Legacy/Shipping/Item_Info.php:228`, and `main_barcode` is a required field with no default at `:434`). Label generation fails for the rest of the run.

On single-order saves (fresh request each time) the check is always false, so the reorder #336 built never fires — every V4 label save still makes the standalone V1 barcode call. Your description says the prefetch flow is the deliberate choice for this task, so that half is fine as intent — but note task 14 already merged as #336, so `barcode_from_label()` and the harvest path are live on main, which is what makes this collide.

Fix that matches your stated intent: build the lazy instance somewhere `barcode_from_label()` doesn't look, so `$v4_services` keeps meaning "explicitly injected":

```php
if ( isset( $this->v4_services['label'] ) ) {
	return $this->v4_services['label'];
}
if ( ! isset( $this->label_v4_memo ) ) {
	$this->label_v4_memo = new V4_Label_Service();
}
return $this->label_v4_memo;
```

Then `barcode_from_label()` stays false until the wiring task deliberately injects the service, and single and bulk saves behave the same. Please add a test pinning it: label flag on, nothing injected → `barcode_from_label()` is false both before and after `label_service()` is called. That also covers this new branch, which currently has no test.

One warning for whoever picks this up later: do **not** fix it the other way, by dropping the `isset` from `barcode_from_label()`. Then the harvest path always runs, `main_barcode` is never set, and *every* label save throws "Barcode is empty!" at `Service::create()`'s first line — the V4 service builds a legacy `Item_Info` before checking eligibility, and its legacy fallback (`create_label_pipeline()`) reads `$post_data['main_barcode']` too.

src/Rest_API/V4/Label/Service.php
```diff
@@ -0,0 +1,316 @@
...
+	public function create( array $post_data ): array {
+		$item_info = new Shipping\Item_Info( $post_data );
+		$signals   = $this->gather_signals( $item_info, $post_data );
+
+		if ( ! Eligibility::is_eligible( $signals ) ) {
+			return $this->create_label_pipeline( $post_data );
+		}
+
+		$fields  = $this->extract_fields( $item_info, $signals['mapped'], $post_data );
+		$request = Request_Builder::build( $fields );
```
tommyshellberg: The `postnl_shipment_addresses` filter never fires on the V4 path. It's applied in `Legacy/Shipping/Client.php:292`, inside legacy request composition — and the V4 path builds its request from `$item_info->shipper/receiver` directly, never constructing that client. A site using the filter (it's a documented third-party extension point since 5.7.0) gets labels printed to the unmodified address, no error, nothing logged. Worse, ineligible orders still fall back to the legacy pipeline and *do* get the filter, so the loss is selective and hard to spot.

I know #352 is the filter carry-forward audit and its branch adds an `apply_filtered_receiver()` call here — but that's on chain B, and this PR is the trunk both chains fork from. If the carry-forward is deliberately deferred to #352, please say so in the description and add it to the flip checklist; right now the deferral isn't recorded anywhere.

src/Rest_API/V4/Label/Service.php
```diff
@@ -0,0 +1,316 @@
...
+				'company'          => $item_info->receiver['company'] ?? '',
+				'first_name'       => $item_info->receiver['first_name'] ?? '',
+				'last_name'        => $item_info->receiver['last_name'] ?? '',
+				'street'           => $item_info->receiver['address_1'] ?? '',
+				'house_number'     => $item_info->receiver['house_number'] ?? '',
+				'house_number_ext' => $item_info->receiver['address_2'] ?? '',
+				'postcode'         => $item_info->receiver['postcode'] ?? '',
+				'city'             => $item_info->receiver['city'] ?? '',
+				'country'          => $item_info->receiver['country'] ?? '',
+				'email'            => $item_info->shipment['email'] ?? '',
```
tommyshellberg, line 303: The merchant's "Shipping confirmation e-mail address" setting is silently dropped on V4. Legacy appends a second contact when it's set (`Legacy/Shipping/Client.php:106-112`, `ContactType '02'` with `Settings::get_shipping_confirmation_email()` — sourced into `shipper['shipping_confirmation_email']` at `Base_Info.php:79`). Nothing on the V4 path reads it.

We checked the SDK: `ShipmentParty` takes a single `Contact` and there's no field for a confirmation contact, so this may be a V4 API-level gap rather than an omission here. But then it's a decision to record, not a silent drop — a merchant who configured that address stops getting PostNL shipping confirmations on V4 orders and nobody is told. Can you confirm with PostNL whether V4 has an equivalent, and either send it or document the loss in the description + flip checklist?

src/Rest_API/V4/Label/Service.php
```diff
@@ -0,0 +1,316 @@
...
+
+	/**
+	 * Build the configured SDK client for the current environment.
+	 *
+	 * @return PostnlClientInterface
+	 */
+	private function build_client(): PostnlClientInterface {
+		$factory = $this->client_factory ?? new Client_Factory( $this->settings );
+
+		$v4_key = method_exists( $this->settings, 'get_api_key_new' )
```
tommyshellberg: Two asks here, both about matching the pattern the checkout twins already settled on in #337/#339.

1. The guarded name is right this time (`get_api_key_new`, same as `has_v4_key()` checks), and when the flow routes to V4 the method must exist — so the `''` fallback can't actually be reached. Which is the argument for deleting the guard: the timeframe and pickup services were both moved to constructor-injected key + `Client_Factory` for exactly this reason (`V4/Timeframe/Service.php` takes them in the constructor). Re-deriving the key from settings behind a `method_exists` is the pattern that hid the `get_v4_api_key` bug for three branches.

2. This service logs nothing, success or failure. The catch block above (line 106) converts and rethrows without a log line, and `new Client_Factory( $this->settings )` here omits the logger, so the SDK stays on its NullLogger. Legacy logs URL, request and response for every label call (`Rest_API/Base.php:227-252`), and `V4/Timeframe/Service` makes the PSR-3 logger a *required* constructor parameter and logs failures with PII-trimmed context. First V4 label incident in production currently has no trail while the merchant's logging toggle looks healthy. Same fix as #339: required logger in the constructor, log the failure in the catch.

src/Rest_API/V4/Label/Service.php
```diff
@@ -0,0 +1,316 @@
...
+
+		foreach ( Response_Mapper::get_labels( $item ) as $label ) {
+			$content = Response_Mapper::decode_content( $label );
+
+			if ( '' === $content ) {
+				continue;
+			}
+
+			// phpcs:disable WordPress.NamingConventions.ValidVariableName.UsedPropertyNotSnakeCase -- Third-party SDK DTO properties.
+			$output_type = null !== $label->outputType ? $label->outputType->value : 'pdf';
```
tommyshellberg, line 357: Question about ZPL: the file extension here comes from the SDK enum, which gives `zpl` — but `Order\Base::merge_labels()` only knows `zpl_rle` (its switch has no `zpl` case and falls through to `return array()`). So a Zebra merchant whose label format is A4 would get `maybe_merge_labels()` building a merged record with no filepath.

What extension does the V1 response actually produce for Zebra output? If legacy files end up as `.zpl_rle`, V4 should either emit the same extension or `merge_labels()` needs a `zpl` case. If Zebra + A4 is a combination that can't occur in practice, a one-line comment saying so would settle it.

src/Rest_API/V4/Label/Request_Builder.php
```diff
@@ -0,0 +1,221 @@
...
+		$label_fields    = $fields['label'] ?? array();
+
+		$sender = ShipmentParty::asSender(
+			address: self::address( $sender_fields )
+		);
+
+		$receiver = ShipmentParty::asReceiver(
+			address: self::address( $receiver_fields ),
+			contact: self::contact( $receiver_fields ),
+			receiverType: ReceiverType::Consumer
```
tommyshellberg, line 81: Every receiver is declared a consumer, including B2B orders with a company name — the SDK enum also has `ReceiverType::Business`. V1 had no such field (it just sent `CompanyName`), so V4 forces a choice legacy never made. Does PostNL treat consumer-vs-business differently (delivery windows, neighbour delivery, business-address handling)? If yes, this should branch on whether the receiver has a company name; if it makes no operational difference, a comment saying "deliberate, PostNL confirmed no behavioural difference" would stop the next reviewer asking.

tommyshellberg (commented):
Base is `main`, mergeable is clean, CI is green on both suites.

**One blocking problem** (first inline comment): the `label_service()` change writes the self-built V4 service into `$v4_services['label']`, which is the array `barcode_from_label()` checks. Once the flag is on, the first order of a bulk run works and every later order in the same run fails with "Barcode is empty!". Details and a fix that matches the barcode decision you documented in the description are in the comment.

**Two silent drops on the V4 path** a merchant or third party would only find in production: the `postnl_shipment_addresses` filter never fires, and the "shipping confirmation e-mail" setting is never sent. Inline comments with evidence.

**Mutation testing** (this is the gate that caught the vacuous fixtures in #337 and #339): 21 mutations, 15 caught, 6 survived. The good news first — `Eligibility` is excellently tested: deleting any guard fails a named test, and replacing `is_eligible()` with `return true;` fails 14. `Response_Mapper` is close behind (one gap: malformed base64). All six survivors are in `Request_Builder`'s enum handling and contact mapping, and four of them are the same trap as the last two PRs: the fixture value equals the code's fallback, so the test passes even if the mapping is deleted. Inline comments name the exact fixture changes.

**Checked and fine, so you don't need to defend these:**
- `composer.json` untouched; SDK stays in require-dev.
- Routing: `has_v4_key()` + `postnl_sdk_enable_label` filter, both off by default. Verified the guarded method names match what #309 ships.
- Label record shape matches legacy (`type`/`barcode`/`created_at`/`filepath`) with `api_version` added on top — additive, safe.
- No changes to `Order/`, `Frontend/`, `Checkout_Blocks/`. The `barcode_from_label()` definition and its call site are both present on this branch.
- Weight is genuinely grams end to end (`Utils::maybe_convert_weight()` → `Dimensions(weightGr:)`).
- Address mapping matches legacy for both parties, including the sender house-number-in-address_2 quirk. The `postnl_order_weight` filter still fires (it lives inside `Item_Info`, which the V4 path reuses).
- Eligibility fails toward legacy on every input we could construct — insured, letterbox 24/48h, ID check, signature, delivery-code combos, pickup, multi-collo, Evening, Canary Islands. No case found where a non-plain parcel slips through to V4.
- `write_label_file()`'s skip-if-exists mirrors legacy `put_label_content()` exactly, and V4 is stricter than legacy about not persisting a record for a file that failed to write.
- PHPCS: zero errors/warnings on the five changed src files with the project standard.

**Smaller items:** the four new files say `@since 5.9.9` while the #336/#337 code on main says `@since 6.0.0` — pick one (6.0.0 matches the rest of the migration). `Service.php` itself has no unit tests (it's WooCommerce-bound, so noted, not blocking), and the new lazy-build branch in `label_service()` is also untested — the test suggested in the first comment would cover it.

**Flip-checklist note:** the flip checklist records the barcode-ordering issue as S7 and suggested dropping the `isset` from `barcode_from_label()`. That suggestion is wrong for this branch — see the first comment for why — and the checklist has been corrected.

tests/php/unit/Rest_API/V4/Label/Request_BuilderTest.php
```diff
@@ -0,0 +1,190 @@
...
+	private function domestic_fields(): array {
+		return array(
+			'sender'        => array(
+				'company'          => 'My Shop',
+				'street'           => 'Siriusdreef',
+				'house_number'     => '42',
+				'house_number_ext' => '',
+				'postcode'         => '2132WT',
+				'city'             => 'Hoofddorp',
+				'country'          => 'NL',
```
tommyshellberg, line 39: Every enum fixture in this file sits exactly on the code's fallback value, so the tests pass even if the mapping is deleted. We mutation-tested it: hardcoding `country()` to always return `Country::NL`, `output_type()` to `PDF`, `resolution()` to `DPI_200`, and `shipment_type()` to `Parcel` — all four survive the suite, because the fixtures are `'NL'`, `'pdf'`, `200`, `'parcel'`. This is the same fixture-equals-fallback trap from #337 and #339.

Concrete fixes, one fixture per resolver on a non-fallback value:
- a receiver country of `'be'` (lowercase) asserting `'BE'` — pins both the mapping and the `strtoupper`, which is also currently unpinned;
- `output_type => 'zpl'` asserting `'zpl'` (right now a merchant with a Zebra printer silently getting PDFs would not fail any test — `printer_type_to_label_settings` extracts `'zpl'` correctly and is well tested, but nothing proves `build()` preserves it);
- `resolution => 600` asserting `600`;
- `shipment_type => 'letterbox'` asserting `'letterbox'`.

Two more from the same pass: swapping `firstName`/`lastName` in `contact()` survives — `'Jan'`/`'Jansen'` are in this fixture but never asserted, so reversed names on every label would go unnoticed; add the two assertions. And an invalid country (e.g. `'XX'`) asserting the `'NL'` fallback would pin the fallback branch itself, which no test currently executes.

For balance: `Eligibility` came through the same mutation pass perfectly — every deleted guard fails a named test, and `is_eligible() { return true; }` fails 14. That's the strongest suite in the migration so far.

tests/php/unit/Rest_API/V4/Label/Response_MapperTest.php
```diff
@@ -0,0 +1,146 @@
...
+		$this->assertSame( array(), Response_Mapper::get_labels( new ShipmentShippingItem() ) );
+	}
+
+	/**
+	 * @testdox decode_content() base64-decodes the label document bytes.
+	 */
+	public function test_decode_content_decodes_base64(): void {
+		$label = new Label( label: base64_encode( 'PDF-BYTES' ), outputType: LabelOutputType::PDF );
+
+		$this->assertSame( 'PDF-BYTES', Response_Mapper::decode_content( $label ) );
```
tommyshellberg, line 117: One gap: flipping `base64_decode`'s strict flag to `false` survives the suite. Only valid-base64 fixtures exist, and the two modes only disagree on malformed input — strict returns `false` (so `decode_content()` returns `''` and the label is skipped), non-strict silently decodes garbage and the service writes a corrupt file to disk as a "label". Add a malformed fixture, e.g. `new Label( label: '!!!not-base64!!!' )`, asserting `''`.

(The rest of this file is solid — barcode capture, the empty-label filter, and the record shape all failed their mutations, and the fallback test correctly uses a distinct `'fallback-barcode'` value rather than something the code would produce anyway.)

src/Rest_API/Service_Factory.php
tommyshellberg: Fixed in 1f388ed. The self-built service now lives in a separate `$label_v4_memo` property, so `$v4_services` only ever holds explicitly injected services and `barcode_from_label()` stays false until the wiring task injects one. New tests in `Service_FactoryTest`: `test_label_service_call_does_not_flip_barcode_from_label` (this is the one that failed before the fix), plus memoisation, V4-instance and injected-double-wins tests. Mutation-checked both ways: making `barcode_from_label()` always return true fails 4 tests, always false fails 2 different ones, and putting the self-build back into `$v4_services` fails the flip test.

src/Rest_API/V4/Label/Service.php
tommyshellberg: Fixed in 9997824, mirroring the Timeframe/Pickup pattern: the constructor now requires `Client_Factory`, the resolved key (`#[\SensitiveParameter]`) and a PSR-3 logger. The `method_exists` guard and the settings re-read are gone — `build_client()` is a one-liner over the injected key. Failures are logged before the rethrow with trimmed context (order reference + country + first 4 postcode characters; no names, streets, emails or full postcodes). `Service_Factory` passes the key from `get_api_key_new()` and a `Logger_Adapter`-wrapped logger. Four new tests in `V4/Label/ServiceTest`; deleting the log call, un-trimming the context and blanking the key each fail different named tests.

tests/php/unit/Rest_API/V4/Label/Request_BuilderTest.php
tommyshellberg: Fixed in 7a377fd. Every suggested fixture added: lowercase `'be'` asserting `'BE'` (also pins the strtoupper), `'XX'` asserting the `'NL'` fallback, `'letterbox'`, a zpl/600 + jpg/300 provider for output type and resolution, and Jan/Jansen assertions in the domestic payload test. Re-ran all six mutations plus two extras (delete strtoupper, change the fallback to BE) — every one now fails a named test. All assertions read the output of `build()` via `toArray()`, not the input fixture.

tests/php/unit/Rest_API/V4/Label/Response_MapperTest.php
tommyshellberg: Fixed in 7a377fd: added `test_decode_content_rejects_malformed_base64` with a `'!!!not-base64!!!'` label asserting `decode_content()` returns `''`, plus a guard assertion that non-strict decoding *would* produce bytes (so the test can't pass vacuously). Flipping the strict flag to false now fails it.

src/Rest_API/V4/Label/Service.php
tommyshellberg: Confirmed and fixed in 2b33c58 — this was reachable on this branch, not deferred to multi-collo: eligibility doesn't look at printer type or label format, and `maybe_merge_labels()` only skips merging for a single label when the format is A6. So Printer Type = ZPL + Label Format = A4 + one domestic parcel hits `merge_labels()` with a `zpl` extension, matches no case, and produces a merged record with no filepath. Fix: a `zpl` case falling through to the existing text merger — strictly additive, since nothing produces `zpl` today except the V4 path.

On what V1 actually returns: we couldn't verify it against any published source. PostNL's spec and their official Python SDK type `OutputType` as a free string with only `PDF` as an example, and a GitHub-wide search for `ZPL_RLE` finds only this plugin — the `zpl_rle` case came from one observation in #184. If V1 actually returns `ZPL` or `Zebra`, the legacy merge has the same bug and `zpl_rle` is dead code. One sandbox label with Printer Type = ZPL settles it; added to the flip checklist's sandbox verification list.

src/Rest_API/V4/Label/Service.php
tommyshellberg: Recorded as a deliberate deferral to #352 — in the PR description and in the flip checklist (S15), which is what was missing. No code change here, for a concrete reason beyond ownership: #352's hunk is already stale against this branch (its `create()` conflicts with the `confirm_label()` extraction from 9997824, and its integration test constructs the service with no arguments, which is now an `ArgumentCountError`), so #352 has to rebase and re-apply regardless — a second implementation on this trunk would just give that rebase two variants to resolve. Nothing reaches merchants in the meantime (both gates default off), and the pre-flip verification list already requires proving the filter fires before the label flag goes on. Scope note recorded too: under this PR's eligibility the filter's array can only ever hold the AddressType '01' entry, so #352's single-entry carry-forward is complete for this stage.

src/Rest_API/V4/Label/Service.php
tommyshellberg: Verified: the SDK genuinely cannot express it. `ShipmentParty` takes one nullable `Contact`; `Contact` has one `email`; the request's `toArray()` emits a fixed key list with no contacts block; `PayloadKey` has 167 cases and `email` is the only email-shaped one; and the normalizer only serializes reflected `#[Field]` attributes, so there's no way to inject an extra key without patching the vendored SDK. So this is an API-level gap as far as the SDK shows, not a wiring omission.

Recorded as a decided limitation in the PR description and the flip checklist (S14, now a pre-flip gate). A drafted question for PostNL is in the checklist (§7 Q11): does V4 labelconfirm accept a merchant confirmation address at all; if it exists API-side, which SDK release exposes it; if not, has the behavior moved or is it discontinued — and what did ContactType '02' actually do on V1. If the answer is 'no equivalent', the setting's description gets marked V1-only before the flip so merchants aren't left believing it still works.

src/Rest_API/V4/Label/Request_Builder.php
tommyshellberg: Left as-is for now, with the question queued for PostNL — we checked and the semantics are unknowable from code. The SDK's `ReceiverType` is a bare two-case enum; nothing in the SDK branches on it, validates it, or documents what `business` changes. Worth noting two things: V1 has no equivalent field at all (it just sends `CompanyName`), so `Consumer` is a newly invented default rather than a carry-forward; and the field is nullable — passing null omits the `type` key entirely, which is the closest match to what V1 sends today. The company name itself isn't lost (it goes out on both `Address.companyName` and `Contact.companyName`); only the discriminator is fixed.

Question for PostNL (flip checklist §7 Q12): what does `receiver.type` change operationally; what happens when it's omitted; and is `business` implied by a company name or by an account contract. If the answer is 'it matters', the change is a one-line branch on a non-empty company name plus a non-vacuous business-case fixture; if 'omit it', we pass null and assert the key is absent. The current value is pinned by `Request_BuilderTest`, so any later change is a visible test edit, not drift.

tommyshellberg:
**Fixed:**
- 1f388ed — the bulk-crash: `label_service()` self-builds into a separate memo, so `barcode_from_label()` stays false until a service is deliberately injected (inline reply on the Service_Factory thread has the test/mutation detail).
- 9997824 — constructor-injected key + required PSR-3 logger, matching the Timeframe/Pickup pattern; failures now logged with trimmed context.
- 7a377fd — all six surviving mutants killed (non-fallback fixtures, name assertions, malformed-base64 strict test).
- 2390ada — `@since 5.9.9` → `6.0.0` in the four V4/Label files. Three 5.9.9 stragglers remain in `SDK/Client_Factory`, `SDK/Logger_Adapter` and the `Service_Factory` class docblock, but they're byte-identical to main — separate cleanup, kept out of this diff.
- 2b33c58 — `merge_labels()` gained a `zpl` case; this was reachable on this branch (ZPL printer + A4 format + single parcel), detail on the inline thread.

**Decided, no code:** the `postnl_shipment_addresses` carry-forward is deferred to #352 and the ContactType '02' confirmation e-mail is recorded as a decided limitation (the SDK cannot express it) — both now in the PR description and the flip checklist, with the PostNL questions drafted. `ReceiverType::Consumer` stays pending PostNL's answer on what the field does.

The description also gained a staleness note: the base is main and task 14 (#336) is merged, so the 'stacked PR' paragraph and the barcode 'decision to review' reflect the pre-merge state; the pre-issued-barcode flow remains this PR's behavior until the wiring task injects the label service.

PR #344, opened 2026-07-21
19. V4 shipping label — service flags

sal4sup:
### All Submissions:

* [ ] Have you tested both blocks/non-blocks cart/checkout? _(N/A — admin label generation, not checkout.)_
* [ ] Have you tested the cart and checkout as both a logged-in user and a guest? _(N/A)_
* [ ] Have you successfully placed an order as both a logged-in user and a guest? _(N/A)_
* [ ] Did you have the query monitor plugin active during all testing? _(N/A — see testing note)_

ClickUp task 19: https://app.clickup.com/t/868jqaqh2 (V4 SDK migration; parent https://app.clickup.com/t/868jpv52c). **Stacked on #343 (task 18)** — merge that first; GitHub retargets this to `main` afterwards.

Adds the discrete delivery **Services** on top of the task-18 core label, still behind the two gates (validated V4 key + `postnl_sdk_enable_label`, default off).

- `Request_Builder` — builds the SDK `Services` DTO from resolved flags: `deliveryConfirmation` (signature / deliverycode), `insuredValue`, `statedAddressOnly`, `returnWhenNotHome`, `minimalAgeCheck` (16+/18+), and their combinations. The block is omitted entirely when no recognised service is present, so a plain parcel is unchanged.
- `Eligibility` — the gate now **allows** service-bearing parcels (removed the "no services" and product-option gates, since the mapper's `services` array is the authoritative full V4 representation of a combination it confirms). Adds `resolve_services()` to fill the mapper's `<order_total>` insurance placeholder with the order subtotal.
- `Service` — passes the resolved services into the request.

Because the real selected options flow through `V4_Mapper` (task-18 fix), every domestic combination resolves to its exact row: e.g. `signature_on_delivery`→3189 (`deliveryConfirmation=signature`), `only_home_address`→3385 (`statedAddressOnly`), `return_no_answer`→3090 (`returnWhenNotHome`), `delivery_code_at_door+insured`→3085 (`deliverycode`+`insuredValue`), 3096→signature+stated+return. Combinations with **no** V4 equivalent (e.g. id_check/3438) and anything non-parcel stay on legacy.

**Deferred (kept on legacy, no regression):** pickup-at-PostNL-location variants (codes 3533/3534/…) need `DeliveryLocation` with a PostNL `pickupLocationId`; the legacy path sends the pickup **address**, not a location code, so the correct code source needs confirming before wiring `DeliveryLocation`. Landing as a focused follow-up. International / multi-collo / returns remain tasks 20–21 as before.

1. `npm run test:php:unit` — all green, incl. new `Request_BuilderTest` service cases (each flag + a combination, omitted-when-empty, unknown-enum omitted) and `EligibilityTest` (service-bearing parcel eligible; delivery-code+insured eligible with both services; pickup-location still falls back; `resolve_services` fills the insurance placeholder).
2. Regression (flag off = default): generate a label with signature/insured/etc. selected on a domestic NL order — identical to before (legacy path).
3. Staging (needs the New-API-Key PR + a V4 sandbox key, then `add_filter('postnl_sdk_enable_label','__return_true')`): generate labels for a domestic NL parcel with each service and a combination; confirm the labelconfirm request carries the expected `services` and the label matches a V1 baseline.

### Other information:

### Changelog entry

No changelog entry: internal, feature-flagged, no merchant-facing behavior change (flags default off).


sal4sup:
**Self-review applied (correctness + standards).** No 🔴/🟠. The reviewer confirmed the gate relaxation is sound — the mapper is authoritative for the NL/NL rows it routes to V4, product codes match V1 one-to-one, the only product option on a V4 row (`004/020` delivery-code) is faithfully mapped to `deliveryConfirmation=deliverycode`, and the insured amount matches legacy exactly (`get_subtotal()`).

Addressed the two 🟡 notes:
- **Insured-amount clarity:** added a comment in `resolve_services()` stating the value must stay the item subtotal (legacy `Amounts` parity) and must not become the order total.
- **Acceptance coverage ("every domestic option"):** added a data-provider test over all 11 NL→NL `has_v4_equivalent` parcel rows, asserting each routes to V4 and yields exactly the expected resolved services (base, stated-address, return-when-not-home, signature, and every documented combination incl. 3094/3096 and delivery-code+insured).

Informational (no change): `minimalAgeCheck` support is forward-looking — no NL row emits it yet (id_check combinations are `needs_confirmation` → legacy); insured with a €0 subtotal emits `insuredValue: 0.0`, matching legacy.

Full unit suite green (437 tests); PHPCS clean on all changed source.

src/Rest_API/V4/Label/Service.php
```diff
@@ -202,6 +197,10 @@ private function extract_fields( Shipping\Item_Info $item_info, array $mapped, a
 			'weight_gr'     => (int) ( $item_info->shipment['total_weight'] ?? 0 ),
 			'reference'     => (string) ( $item_info->shipment['order_number'] ?? '' ),
 			'barcode'       => (string) ( $post_data['main_barcode'] ?? '' ),
+			'services'      => Eligibility::resolve_services(
+				$mapped['services'] ?? array(),
+				(float) ( $item_info->shipment['subtotal'] ?? 0 )
```
tommyshellberg, line 307: Nothing tests this line. Changing it to `0` leaves the whole suite green — no test anywhere instantiates `V4\Label\Service`, because `extract_fields()` and `gather_signals()` are private and the class extends `Order\Base`.

That matters because this is the one number that reaches PostNL as the insured amount, and the docblock on `resolve_services()` has to warn in prose: "Do not switch it to the order total — that would add tax/shipping and diverge from V1." A warning in a comment is the weakest possible enforcement. Someone will "fix" this to `get_total()` and every insured V4 shipment will be over-declared, with nothing failing.

`Eligibility::resolve_services()` being well tested in isolation doesn't cover this — the untested part is which number gets handed to it.

Cheapest real fix: an integration test alongside the existing ones that builds a real insured domestic order and asserts `services['insuredValue']` equals the item subtotal, with a fixture where subtotal, total and 0 are three different numbers (subtotal 42.00, shipping 6.95, total 48.95). Then neither `0` nor `get_total()` passes. Failing that, pull `extract_fields()` out into a pure helper the way `Request_Builder` and `Eligibility` already are.

src/Rest_API/V4/Label/Request_Builder.php
```diff
@@ -101,10 +110,42 @@ public static function build( array $fields ): ShipmentDeliveryRequest {
...
+	 * Translate resolved service flags into a V4 Services DTO.
+	 *
+	 * Returns null when no recognised service is present, so the request omits
+	 * the Services block entirely for a plain parcel.
+	 *
+	 * @param array $flags Resolved service flags keyed as documented on build().
+	 * @return Services|null
+	 */
+	private static function services( array $flags ): ?Services {
+		$confirmation = DeliveryConfirmation::tryFrom( (string) ( $flags['deliveryConfirmation'] ?? '' ) );
```
tommyshellberg, line 131: If this string ever stops matching the enum, the label silently ships without the service the customer paid for.

`tryFrom()` returns null for anything it doesn't recognise. Matrix row 5 (`signature_on_delivery` → 3189) has `deliveryConfirmation` as its *only* service, so if someone writes `'Signature'` with a capital S, or PostNL renames the case in an SDK bump, all five locals go null, `services()` returns null, and the request carries no Services block. V4 sends no product code, so there's no second field carrying the intent. The merchant charged for signature-on-delivery and the courier leaves the parcel on the doormat. No error, no log line.

The values match today — I checked `'signature'` and `'deliverycode'` against the enum, exactly. The gap is that nothing keeps them matching. The literals are hand-copied in three places (`V4_Mapper`, `V4_MapperTest`, `EligibilityTest::domestic_option_provider`) and no test references the enum.

I'd suggest a contract test rather than importing the enum into `V4_Mapper`: assert every `deliveryConfirmation` and `minimalAgeCheck` string the matrix emits is a valid case. Same loud failure, but at CI. Importing `Postnl\Sdk\Enums` into `V4_Mapper` would tie a plain lookup table to a `require-dev` package that production builds strip out, which is the shape of the PHP 8.2 fatal that already shipped once.

tests/php/unit/Rest_API/V4/Label/EligibilityTest.php
```diff
@@ -77,13 +101,11 @@ public static function ineligible_provider(): array {
 			'delivery-day selected'      => array( array( 'is_delivery_day' => true ), 'a delivery-day option' ),
 			'pickup selected'            => array( array( 'is_pickup' => true ), 'a pickup point' ),
 			'return involved'            => array( array( 'has_return' => true ), 'a return label' ),
-			'product options present'    => array( array( 'has_product_options' => true ), 'product options' ),
 			'evening delivery'           => array( array( 'delivery_type' => 'Evening' ), 'evening delivery' ),
 			'non-NL origin'              => array( array( 'origin' => 'BE' ), 'a non-NL origin' ),
 			'non-NL destination'         => array( array( 'destination' => 'BE' ), 'a non-NL destination' ),
 			'no v4 equivalent'           => array( array( 'mapped' => array( 'has_v4_equivalent' => false ) ), 'no V4 equivalent' ),
```
tommyshellberg: This row passes for a different reason than its name says, and as a result `has_v4_equivalent` can be deleted from `is_eligible()` with the suite still green.

`signals()` uses `array_merge`, not a deep merge, so `mapped` is replaced wholesale and `shipmentType` goes missing. `'parcel' === ($mapped['shipmentType'] ?? '')` rejects it before `has_v4_equivalent` is ever read.

On real mapper output the two are correlated — `legacy_result()` never emits `shipmentType` — so this isn't a live bug. But `has_v4_equivalent` is the flag keeping every `needs_confirmation` and `NOT_YET_AVAILABLE_CODES` row off V4, and it's the one gate in this function with no test that actually exercises it. Make the row identical to the happy path except for the flag under test:

```php
'no v4 equivalent' => array( array( 'mapped' => array( 'has_v4_equivalent' => false, 'shipmentType' => 'parcel', 'services' => array(), 'deliveryLocation' => array() ) ), 'no V4 equivalent' ),
```

src/Rest_API/V4/Label/Service.php
```diff
@@ -202,6 +197,10 @@ private function extract_fields( Shipping\Item_Info $item_info, array $mapped, a
 			'weight_gr'     => (int) ( $item_info->shipment['total_weight'] ?? 0 ),
 			'reference'     => (string) ( $item_info->shipment['order_number'] ?? '' ),
 			'barcode'       => (string) ( $post_data['main_barcode'] ?? '' ),
+			'services'      => Eligibility::resolve_services(
```
tommyshellberg, line 305: A store that doesn't sell in euros insures the wrong amount, and this PR is what starts it.

Legacy sends the currency with the value — `Legacy/Shipping/Client.php:178-182` builds `['AmountType' => '02', 'Currency' => $this->item_info->shipment['currency'], 'Value' => ...]`, where currency is `$order->get_currency()`. The V4 `Services` DTO has a bare `?float $insuredValue` and no currency field; I grepped all of `RequestData/V4/` and the only `Currency` is in `InternationalShipment/Customs.php`, which this flow never builds.

So a Dutch store selling in GBP with a £250 insured order declares `Value: 250, Currency: GBP` on V1 and a bare `250` on V4. If PostNL reads that as euros the merchant is roughly 15% under-insured, and finds out at claim time.

This is an API shape question, not a coding mistake — worth asking PostNL whether V4 `insuredValue` is EUR-only. If it is, the honest fix is to refuse V4 when `$order->get_currency() !== 'EUR'` rather than send a number whose unit the request doesn't state. Either way it belongs in the flip checklist before the label flag goes on anywhere.

src/Rest_API/V4/Label/Eligibility.php
```diff
@@ -92,11 +91,38 @@ public static function is_eligible( array $signals ): bool {
 			return false;
 		}
 
+		// The mapper is authoritative: for a combination it marks as having a V4
+		// equivalent, its services array is the full V4 representation of every
+		// selected option, so no separate product-option gate is needed. Only a
+		// pickup DeliveryLocation is excluded here — that variant lands separately.
 		$mapped = $signals['mapped'] ?? array();
 
 		return ! empty( $mapped['has_v4_equivalent'] )
```
tommyshellberg, line 100: Worth knowing about the "mapper is authoritative" argument: it has one hole, and it's the row this PR adds.

The backstop when option flags and the resolved product disagree is `V4_Mapper`'s `REASON_PRODUCT_CODE_MISMATCH` check. That works for rows 3-11, which all have distinct codes. It can't work for row 2: `delivery_code_at_door + insured_shipping` maps to product **3085**, the same code as row 1, the plain base parcel. V1 carried the difference in a product option (`004`/`020`) plus an Amounts block, not in the code.

So if the options key ever collapses to `'(base)'` for an order that genuinely has both flags, row 1 matches, the product code matches, `is_eligible()` returns true, and the parcel ships with no delivery code and no insurance while the order screen says otherwise.

I could not build a live path to this and neither could a second pass — the meta box writes `'yes'` and `save_meta_value()` deliberately reads the untrimmed `save` context. So this is hardening, not a defect. But it's the exact shape of review trigger #2 in `agents.md`, and the gate that used to catch it is the one this PR removes. An assertion that a row sharing the base row's product code must carry non-empty services would close it permanently.

src/Rest_API/V4/Label/Request_Builder.php
```diff
@@ -101,10 +110,42 @@ public static function build( array $fields ): ShipmentDeliveryRequest {
...
+		$age_check    = MinimalAgeCheck::tryFrom( (string) ( $flags['minimalAgeCheck'] ?? '' ) );
```
tommyshellberg, line 132: `minimalAgeCheck` can never be set. I read all 89 rows of the matrix — none emits the key. Every combination that would need one is an `id_check` row, and all of them are `legacy_result` (rows 13-20 and 23-24 for NL→NL, row 72 for BE→NL), so `tryFrom()` here only ever receives `''`.

That's the right call for behaviour — ID Check should stay on legacy until it's confirmed. The problem is that the code says otherwise. This line, the `Services` constructor call, the `build()` docblock ("minimalAgeCheck ('16+'|'18+')"), the `resolve_services()` docblock, and two `service_flag_provider` rows all read as though 18+ orders route through V4. The PR description says the same. The next person planning the ID Check migration will think it's done.

A one-line comment saying no matrix row produces this yet is enough. The ClickUp task does ask for 18+/16+, so it's worth being explicit in the description that this part is handling-only, not reachable.

src/Rest_API/V4/Label/Eligibility.php
```diff
@@ -92,11 +91,38 @@ public static function is_eligible( array $signals ): bool {
...
 			&& 'parcel' === ( $mapped['shipmentType'] ?? '' )
-			&& empty( $mapped['services'] )
 			&& empty( $mapped['deliveryLocation'] );
 	}
+
+	/**
+	 * Resolve the mapper's service placeholders into concrete request values.
+	 *
+	 * The matrix stores insuredValue as the '<order_total>' placeholder; it is
```
tommyshellberg: The placeholder is called `<order_total>` but the value substituted is the item subtotal, which is why this docblock has to spend three lines telling the reader not to "fix" it. Renaming it to `<order_subtotal>` in the matrix costs nothing and removes the trap instead of documenting it.

src/Rest_API/V4/Label/Request_Builder.php
```diff
@@ -101,10 +110,42 @@ public static function build( array $fields ): ShipmentDeliveryRequest {
...
+		// isset (not ??) so a missing insured value stays null instead of coercing to 0.0.
```
tommyshellberg: This comment names the wrong comparison. `??` handles a missing key exactly the same way `isset` does here — both give null. The distinction that actually matters is `isset` vs `! empty`, which is what lets `insuredValue => 0.0` reach the request instead of being dropped. Mutation testing confirms the code is right and `test_zero_insured_value_is_sent` catches the `! empty` swap; only the comment is off.

tommyshellberg (changes requested):
Reviewed against `origin/main` at `388020b` — #343 landed while I was reading this.

The service-flag work itself is good. The mapper-driven approach is sound, the tests are better than most in this queue, and I could not find a case where a shipping service gets silently dropped. The inline comments below are what I'd like changed.

## Mutation testing

**9 of 16 caught** (measured pre-merge). What's already solid:

- `return null` from `services()` kills 9 tests.
- Swapping the `statedAddressOnly` and `returnWhenNotHome` constructor arguments is caught. Both are `bool|null`, so this is the classic swap that normally ships green — it's caught because `test_services_combination_is_mapped` asserts `assertArrayNotHasKey('statedAddressOnly', ...)` next to a set `returnWhenNotHome`. Nice test.
- `isset` → `! empty` on a `0.0` insured value is caught by `test_zero_insured_value_is_sent`. That is exactly the kind of thing that ships silently and you wrote the test for it.
- The `<order_total>` substitution uses 49.95 and 42.0, so it isn't testing against its own default.
- `domestic_option_provider`'s 11 rows are exactly the 11 NL→NL rows in `V4_Mapper` with a V4 equivalent. Full coverage of the matrix, confirmed by reading it.

One survivor is an equivalent mutant, not a test gap: `array_key_exists` → `isset` in `resolve_services()`. `V4_Mapper` only ever emits the string `'<order_total>'` for `insuredValue`, never `null`, so nothing can tell the two apart. Keep `array_key_exists` as written.

The two remaining survivors are inline below and still need fixing.

## Checked and fine

- **Removing the `has_product_options` gate is safe.** I traced all three sources it covered. Of the 11 NL→NL rows that map to V4, ten have no product options; the one exception is `delivery_code_at_door + insured_shipping` (3085, characteristic `004` / option `020`), which the matrix translates to `deliveryConfirmation: deliverycode`. Every row carrying `002/014` (ID Check) is legacy-only. `get_product_options()` only ever fires for Evening or 08:00-12:00, which `is_eligible()` already rejects twice. Nothing is dropped.
- **The enum strings match.** `DeliveryConfirmation` backs `'signature'` and `'deliverycode'`; `MinimalAgeCheck` backs `'16+'` and `'18+'`. Exact, case included.
- **The `product_code_mismatch` guard is real, not decorative** — it catches letterbox 48 (2948 vs 2928), ID Check (3438 vs 3085), and unrecognised flags falling back to base.
- **The insured gate matches legacy.** V4 emitting `insuredValue` always implies legacy would have emitted `Amounts`.
- **No settings or filters lost beyond `postnl_shipment_addresses`**, which is already a known #343 item. The three `postnl_order_weight` hooks and every `Settings::` read fire inside `Shipping\Item_Info`, which the V4 path also builds, so they're inherited. The €5000 insurance cap and the mailbox weight limit both still apply.
- **`composer.json` untouched** — SDK still `require-dev`, pinned `dev-main#5a064dd`, `php >= 7.4` in `require`.
- No filter or action signature changes, no order meta shape changes, and the PR touches only three `V4/Label` files, so there is no way for it to reach a legacy path.

src/Rest_API/V4/Label/Service.php
tommyshellberg: Fixed in c8aecd5. `ServiceTest` now reaches `extract_fields()` by reflection (rather than promoting it to protected — it has no collaborators) with a fixture where subtotal 42.00, weight 1500 and 0 are three distinct numbers, so reading the wrong key cannot coincidentally pass. Mutation-verified: insured amount to `0` fails it, and so does swapping to `total_weight`.

src/Rest_API/V4/Label/Request_Builder.php
tommyshellberg: Fixed in c8aecd5. `V4_MapperTest` now walks all 89 rows and asserts every emitted `deliveryConfirmation`/`minimalAgeCheck` string is a value the SDK enum accepts. Went with a contract test rather than referencing the enum from `V4_Mapper`, since the SDK is `require-dev` and stripped from production builds — same loud failure, at CI, with no production coupling. Mutation-verified: changing row 5 to `Signature` fails 5 tests.

tests/php/unit/Rest_API/V4/Label/EligibilityTest.php
tommyshellberg: Fixed in c8aecd5. The row now matches the eligible happy path except for the flag under test. Mutation-verified: deleting the `has_v4_equivalent` gate now fails, where it previously left the suite green.

src/Rest_API/V4/Label/Request_Builder.php
tommyshellberg: Addressed in 8d53394 and c8aecd5. Both docblocks now say `minimalAgeCheck` is accepted but that no matrix row emits it, so ID Check does not reach V4 at all. c8aecd5 also adds a test pinning the count of emitting rows at zero, so promoting an id_check row without confirming the age-check mapping becomes a deliberate test edit rather than a silent change.

src/Rest_API/V4/Label/Eligibility.php
tommyshellberg: Partly addressed in 8d53394 — the docblock now names `<order_total>` as a known misnomer instead of leaving the next reader to infer it from the warning underneath. Held off on the actual rename: it touches `V4_Mapper` in seven places plus two test files, which is more churn than belongs in this PR. Worth doing as a standalone commit.

src/Rest_API/V4/Label/Request_Builder.php
tommyshellberg: Fixed in 8d53394. The comment now names isset vs `! empty`, and says why it matters: a legitimately zero insured value still gets sent, matching the legacy Amounts block.

src/Rest_API/V4/Label/Service.php
tommyshellberg: Leaving this open deliberately — it needs an answer from PostNL, not a code change. Recorded as Q13 and W18 in the flip checklist as a **pre-flip gate**, not a pre-merge one: both gates default off, so nothing reaches a merchant when this merges. If PostNL confirm `insuredValue` is EUR-only, the fix is a currency check in the eligibility gate, and doing that before more label PRs stack on top is cheaper than after.

src/Rest_API/V4/Label/Eligibility.php
tommyshellberg: Leaving this open as a tracked note rather than a blocker. Recorded as W20 in the flip checklist. No live path exists today — two independent passes tried to construct one and could not, since the meta box writes `yes` and `save_meta_value()` reads the untrimmed `save` context. Worth an assertion that a matched row sharing the base row's product code must carry non-empty services, but that is hardening and belongs on its own.

tommyshellberg (approved):

PR #347, opened 2026-07-24
20. V4 Shipping label — international (EU/ROW) + customs

sal4sup:
### All Submissions:

ClickUp: [20. V4 Shipping label — international + customs](https://app.clickup.com/t/868jqaqhb)

Extends the V4 SDK label service (tasks 18–19) to EU and ROW shipments, including customs. Stacked on `postnl-v4-shipping-label-service-flags`.

**Outstanding question §2 (`services.bundle`) resolved:** the V4 `Services` DTO has no `bundle` property, but `InternationalShipmentData->bundle` (a `Bundle` enum: `track_trace` / `insured` / `insured_plus`) does — which is the correct home for an international service bundle. So EU/ROW parcels map to V4 via `InternationalShipmentData`, not a `Services` workaround. The `SDK_SERVICES_BUNDLE_GAP` marker is kept as a documented fact about the `Services` DTO.

- **`V4_Mapper`** — flips the 14 international **parcel** rows (4907 EU / 4909 ROW, `delivery_day`) from Legacy-only to V4, each carrying an `internationalShipmentData` bundle hint (base/track&trace → `track_trace`, insured → `insured`, insured_plus → `insured_plus`). Packet/mailbox products and international pickup stay on legacy.
- **`Request_Builder`** — builds `InternationalShipmentData` (bundle + `Customs`), `Customs` (currency, `transactionCode` 11, invoice `associatedDocument`, `senderIdentification`, per-line `Content[]`) and attaches it to the request. Description is truncated to 35 chars; weight/quantity clamped to ≥1; unknown enum values (bundle/currency/country) are omitted.
- **`Service`** — sources customs content from the parsed order line items (description, qty, weight in grams, value, HS code, country of origin), currency from the order, and `senderIdentification` from the merchant code (empty for EU). Mirrors the legacy Customs block.
- **`Eligibility`** — now accepts an NL/BE → EU/ROW parcel in addition to the domestic NL parcel; NL↔BE cross-border still routes to legacy.
- **`Response_Mapper`** — captures the international `partnerBarcode` / `partnerId` from the `labelconfirm` response into the stored label record (no barcode endpoint is called; see task 14).

Both gates (validated New API Key + the `postnl_sdk_enable_label` flag) default off, so merging changes nothing for merchants.

1. `composer install` then run the PHP unit suite: `npm run test:php:unit` (or `vendor/bin/phpunit -c phpunit-unit.xml.dist`) — all pass.
2. `composer check-php` — no violations on the changed files.
3. With a validated New API Key and the label flow flag enabled on a NL/BE store, generate a shipping label for an order to an EU destination (e.g. DE) and an ROW destination (e.g. US): confirm the request carries `internationalShipmentData` with the expected `bundle` and a `customs` block (one `content` entry per line item with HS code, country of origin, value, weight), and that the returned partner barcode is stored in the order meta.
4. Confirm a domestic NL order still produces a label with no `internationalShipmentData` block, and that a packet/mailbox or pickup international order falls back to the legacy path.

### Other information:

### Changelog entry

Internal (no merchant-visible change): add V4 SDK EU/ROW international shipping label generation with customs declaration and partner-barcode capture, behind the default-off `postnl_sdk_enable_label` flag.


src/Rest_API/V4/Label/Service.php
```diff
@@ -370,7 +437,7 @@ private function store_labels( LabelConfirmResponseInterface $response, $order,
 				continue;
 			}
 
-			$records[] = Response_Mapper::to_label_record( $label_type, $barcode, $filepath );
+			$records[] = Response_Mapper::to_label_record( $label_type, $barcode, $filepath, $partner_barcode, $partner_id );
```
tommyshellberg, line 514: `maybe_merge_labels()` throws away `partner_barcode` and `partner_id` for most international shipments — the thing this PR exists to capture.

It returns a record untouched only when there is exactly one label document and the format is A6 (`Order/Base.php:1138`). Every other path builds a brand-new record containing only `type`, `barcode`, `created_at`, `filepath`, `merged_files` (`Order/Base.php:1165`). A ROW response returns four documents (`cn23`, `cp71`, `label`, `commercialinvoice` per `Mapping::label_type_list()`), so ROW always takes the merge path, and an EU label on A4 format does too. The partner barcode reaches order meta only for EU + A6. No error, no log — the labelconfirm response had the barcode and it's just gone.

The test passes because `test_partner_refs_captured_from_response` asserts the record before the merge.

The fix belongs here, not in the shared `maybe_merge_labels()`: the loop just below already re-attaches `api_version` after the merge, and this method already holds `$partner_barcode`/`$partner_id` — add them in that same loop when non-empty. Then add a test that exercises the merged path (multiple label records in, merged record out) and asserts the partner keys survive.

src/Rest_API/V4/Label/Request_Builder.php
```diff
@@ -114,10 +128,120 @@ public static function build( array $fields ): ShipmentDeliveryRequest {
...
+		$content = self::customs_content( $data['content'] ?? array() );
+
+		if ( empty( $content ) ) {
+			return null;
+		}
+
+		return new Customs(
+			content: $content,
+			transactionCode: self::maybe_null( (string) ( $data['transaction_code'] ?? '' ) ),
+			currency: Currency::tryFrom( strtoupper( (string) ( $data['currency'] ?? '' ) ) ),
```
tommyshellberg, line 187: When one of these `tryFrom` calls misses, the field silently disappears from the customs declaration and nothing logs it.

Three concrete cases:
- The SDK `Currency` enum has only 13 currencies. A store selling in HUF, RON, BGN, CNY (or any other missing one) sends a customs block with no currency. Legacy sent the raw string and let the API reject it loudly.
- `Bundle::tryFrom` at line 152: today the mapper only emits the three valid strings, but a one-character typo in a future mapper row would print an uninsured international label with no signal anywhere.
- `Country::tryFrom` at line 213: product origin meta holding "Netherlands" instead of "NL" drops the country of origin from the declaration.

The PR description does disclose that unknown enum values are omitted — but a merchant debugging a rejected customs label has nothing in the logs to find, where the legacy path logged every request body.

This class is a pure static builder, so the log line doesn't belong here. `Service::extract_international()` and `extract_customs()` have `$this->logger` and see the same raw values — one warning there when a bundle/currency/origin value won't map to the enum would close the gap.

tests/php/unit/Rest_API/V4/Label/Request_BuilderTest.php
```diff
@@ -318,6 +318,131 @@ public function test_unrecognised_service_values_omit_block(): void {
...
+	public function test_international_block_omitted_for_domestic(): void {
+		$this->assertArrayNotHasKey( 'internationalShipmentData', $this->payload( $this->domestic_fields() ) );
+	}
+
+	/**
+	 * A representative EU international field set with a customs declaration.
+	 *
+	 * @return array
+	 */
+	private function international_fields(): array {
```
tommyshellberg, line 333: Both `strtoupper` calls in `Request_Builder` survived mutation testing: deleting either one keeps all 591 tests green, because every fixture is already uppercase — `'EUR'` here, `'NL'` at the `country_of_origin` line, and even the negative-path fixture `'XXX'`. So the case normalization is currently untested.

Either make one fixture lowercase (`'eur'` here and `'nl'` for the origin — each is a one-character change and kills the mutation), or, if the legacy `Item_Info` schemas can provably never emit lowercase values, delete the `strtoupper` calls instead. The one state to avoid is the current one: a defensive transform nothing exercises.

(For the record: the other 10 of 12 mutations were caught, including both eligibility allowlists, the weight clamp, the truncation length, and the row-44 bundle swap — the suite is in good shape.)

src/Helper/Product_Mapper/V4_Mapper.php
```diff
@@ -302,10 +310,10 @@ private static function matrix(): array {
 				),
 				'EU'  => array(
 					'delivery_day'  => array(
-						'(base)'                           => self::legacy_result( 42, '4907', $nc ),
-						'track_and_trace'                  => self::legacy_result( 43, '4907', $nc ),
-						'insured_shipping+track_and_trace' => self::legacy_result( 44, '4907', $nc ),
-						'insured_plus+track_and_trace'     => self::legacy_result( 45, '4907', $nc ),
+						'(base)'                           => self::v4_result( 42, '4907', 'parcel', array(), array(), $track_trace ),
+						'track_and_trace'                  => self::v4_result( 43, '4907', 'parcel', array(), array(), $track_trace ),
+						'insured_shipping+track_and_trace' => self::v4_result( 44, '4907', 'parcel', array(), array(), $insured ),
```
tommyshellberg, line 321: Legacy sent the declared value with this insured row; the V4 request carries no amount at all. Worth confirming against the PostNL docs before this row goes live.

On legacy, an insured EU shipment gets `Amounts[AmountType 02, Currency, Value = subtotal]` (`Legacy/Shipping/Client.php:174-186`). Here the row emits `services: array()`, so `Request_Builder::services()` returns null and `insuredValue` is never set — the only insurance signal in the whole request is `bundle: insured`.

If the international bundles are flat-coverage tiers (which is how PostNL's international insured products have historically worked), this is correct and just deserves a sentence in the docblock saying so. If the API expects a declared value alongside the bundle, coverage is silently wrong. Either way, please record the answer in the flip-checklist next to the existing domestic insured-currency question so it's checked before the label flag flips.

tommyshellberg (changes requested):
The core of this PR is right: every SDK DTO and constructor parameter the new code names exists in the pinned SDK (`InternationalShipmentData(pudo, customs, bundle)`, `Customs`, `Content`, `AssociatedDocument`, `partnerBarcode`/`partnerId` on `ShipmentShippingItem`), the `Bundle` enum has exactly the three values the mapper emits, and the outstanding-question-§2 resolution in the description is correct. Legacy customs parity is solid: transaction code 11, invoice + order id, currency, merchant code as `senderIdentification` (already empty for EU via the `Item_Info` sanitizer), one content entry per line. Legacy never sends `HandleAsNonDeliverable`, `receiverIdentification`, or `labelSignature`, so leaving them null loses nothing. Every key `extract_customs()` reads exists in the legacy `Item_Info` schemas, with `description` already truncated to 35 and weight already floored to 1 gram there. NL↔BE cross-border correctly stays on legacy (`get_shipping_zone()` returns `'BE'`/`'NL'`, which the gate rejects), and every international combination the mapper doesn't promote (packet/mailbox, pickup, id_check, unknown combos) falls back to legacy safely.

Mutation testing: 12 mutations, 10 caught. Deleting the international block, hardcoding the transaction code, dropping the weight clamp, breaking the truncation length, dropping the associated document, breaking either eligibility allowlist, blanking the partner barcode, dropping the empty-string guards on the partner keys, and swapping row 44's bundle all fail named tests — and the two eligibility allowlists are pinned by disjoint provider rows, so neither could be replaced by `return true`. The two survivors are both the `strtoupper` calls; see the fixture comment.

One blocking finding: the partner barcode is dropped on the way into order meta for most international shipments (inline comment on `Service.php`). The rest are a logging gap and two questions to settle.

One minor divergence, no comment needed unless you disagree: an EU/ROW order whose line items are all virtual produces `contents = []`, so `customs()` returns null and the request goes out with a bundle but no customs block at all, where legacy sent a customs block with an empty `Content[]`. Probably unreachable in practice (an all-virtual order has nothing to ship) and the API will presumably reject it loudly, but if you want parity it is a one-line change to build the customs block from the shipment fields even when content is empty.

Gates for the record: SDK still in `require-dev` pinned `#5a064dd`, `Service_Factory` and `Order/Base` untouched, `barcode_from_label()` definition and call site both present, label records only gain optional keys, base is `main`, PHPCS 0 errors on the changed files, suite green (591 tests).

src/Rest_API/V4/Label/Service.php
tommyshellberg: Fixed in c3a5492. `store_labels()` now runs the merged records through a new private `finalize_label_records()`, which re-adds `api_version` plus `partner_barcode`/`partner_id` when non-empty — same only-add-when-non-empty rule as `to_label_record()`, and `maybe_merge_labels()` is untouched. Two new tests feed it the exact rebuilt shape from `Order/Base.php:1165`: reverting the re-attach fails `test_partner_references_survive_the_label_merge`, and dropping the empty-string guards fails `test_absent_partner_references_add_no_partner_keys` — different tests for each direction, so neither is vacuous.

src/Rest_API/V4/Label/Request_Builder.php
tommyshellberg: Fixed in c0429e4. `Service::extract_international()` now logs one warning per non-empty value that will not map — bundle, currency, and each content item&apos;s country of origin — naming the field, the raw value, and the order number (same reference `confirm_label()` logs). Each check mirrors the builder&apos;s normalization exactly (`Currency::tryFrom(strtoupper(...))` etc.), so a lowercase `eur` that the builder maps fine does not warn; a dedicated test pins that in both directions (inverting the conditions fails it with warnings for `insured`/`eur`/`nl`). The builder stays a pure static translator — the docblock on `warn_unmappable_international()` explains why the log lives on the service.

tests/php/unit/Rest_API/V4/Label/Request_BuilderTest.php
tommyshellberg: Fixed in f2bb49f — the fixture now feeds `eur` and `nl` while the assertions still expect `EUR`/`NL`, so the builder&apos;s case-folding is what makes them pass. Re-ran both mutations: deleting `strtoupper` at `Request_Builder.php:187` and at `:213` each now fail `test_international_customs_is_mapped` (both survived before). Kept the `strtoupper` calls rather than deleting them, since c0429e4&apos;s warning path relies on the same normalization contract.

src/Helper/Product_Mapper/V4_Mapper.php
tommyshellberg: Documented in 4173d45: the comment above the bundle hints now states that V4 carries only the flat bundle enum with no declared value, that V1 sent `Amounts[02, currency, subtotal]` with insured EU shipments, and that flat-tier vs declared-value is an open PostNL question — tracked as Q14 in the flip-checklist and a pre-flip gate for the label flow. No logic change; once PostNL answers, either the docblock becomes the confirmation or the mapper rows gain an amount.

tommyshellberg (approved):

PR #348, opened 2026-07-25
21. V4 Shipping label — multi-collo + PDF storage + response mapping polish

sal4sup:
### All Submissions:

ClickUp: [21. V4 Shipping label — multi-collo + PDF storage + response mapping polish](https://app.clickup.com/t/868jqaqhd)

Extends the V4 SDK label service to multi-parcel (multi-collo) shipments and finalises response-mapping/PDF-storage parity with the legacy V1 path. Stacked on `postnl-v4-shipping-label-international-customs` (task 20).

- **`Request_Builder`** — builds one `ShippingItem` per collo from a `barcodes` array (SDK derives `itemCount` from the item count); each collo shares the order reference and weight, matching the legacy per-shipment payload. Falls back to a single item from `barcode` when no `barcodes` array is present, so single-collo behaviour is unchanged.
- **`Service`** — passes the pre-issued `barcodes[]` through; `store_labels()` now iterates **every** returned shipment item, writes each collo's label with its own barcode, and merges them all via `Order\Base::maybe_merge_labels()`. For >1 collo this produces a single merged sheet keyed by the parent (main) barcode — identical to V1; per-collo `barcodes[]` meta is written by `Order\Base` as before.
- **`Eligibility`** — no longer rejects `num_labels > 1`; only an invalid (`< 1`) collo count falls back.
- **`Response_Mapper`** — adds `all_shipment_items()` to return every collo item in order.

PDF storage parity is inherent: filenames use the same `Utils::generate_label_name()` and merging uses the same `Order\Base::maybe_merge_labels()` the legacy path calls, so filenames, the merged `merged_files[]` record, and the `labels`/`barcodes[]` meta shape are indistinguishable from V1 for downstream consumers (packing slips, T&T emails). Both gates (validated New API Key + `postnl_sdk_enable_label` flag) default off.

1. Run the PHP unit suite: `vendor/bin/phpunit -c phpunit-unit.xml.dist` — all pass (458 tests).
2. `composer check-php` — no violations on the changed files.
3. With a validated New API Key and the label flow flag on a NL store, set an order's "Number of labels" to 3 and generate the label. Expect: the request carries `itemCount: 3` with three `items[]` (one barcode each); three barcodes stored in `_postnl_order_metadata['barcodes']`; a single merged label PDF in `labels['label']` with `merged_files[]` — the same meta shape a V1 3-collo order produces.
4. Confirm a single-collo order still generates exactly as before (one item, one label), and that a multi-collo EU/ROW order also works (combined with task 20).

### Other information:

### Changelog entry

Internal (no merchant-visible change): add multi-collo support to V4 SDK shipping-label generation with legacy-parity PDF storage and response mapping, behind the default-off `postnl_sdk_enable_label` flag.


src/Rest_API/V4/Label/Service.php
```diff
@@ -157,7 +157,9 @@ public function create( array $post_data ): array {
 		$request  = Request_Builder::build( $fields );
 		$response = $this->confirm_label( $request, $fields );
 
-		return $this->store_labels( $response, $post_data['order'], (string) $fields['barcode'] );
+		$barcodes = ! empty( $fields['barcodes'] ) ? $fields['barcodes'] : array( (string) $fields['barcode'] );
```
tommyshellberg, line 160: This only works when the barcodes were pre-issued. On the harvest path — `barcode_from_label()` true, which is exactly the wiring this migration is building toward — `Order\Base` calls `create_label()` with no `barcodes` and no `main_barcode` in `post_data` (`Order/Base.php:629-636`). Then `$fields['barcodes']` is empty, `Request_Builder::items()` falls back to one barcode-less item, and a 3-collo order is sent to PostNL as a single collo. Nothing on the request side ever reads `num_labels`, so nothing notices: the merchant gets one label and one barcode for a shipment they declared as 3 colli, the meta still says 3, and there is no error and no log line.

Suggest either building the items from `num_labels` when no barcodes exist (N barcode-less items), or making multi-collo ineligible on the harvest path until that's supported — either way the mismatch should be loud, not silent. One caveat for the first option: nothing in the SDK source or docs says N barcode-less items produce N auto-issued labels, so that behavior needs a sandbox check before it's relied on. This needs an entry in the flip-checklist regardless of which way you go.

src/Rest_API/V4/Label/Service.php
```diff
@@ -462,26 +465,66 @@ private function build_client(): PostnlClientInterface {
...
 
-		if ( null === $item ) {
+		if ( empty( $items ) ) {
 			throw new \Exception(
 				esc_html__( 'Cannot create the label. No shipment was returned by PostNL.', 'postnl-for-woocommerce' )
 			);
 		}
 
-		$barcode         = Response_Mapper::get_barcode( $item, $fallback_barcode );
+		$parent_barcode = (string) ( $fallbacks[0] ?? '' );
```
tommyshellberg: On main this was the barcode issued on the response (`Response_Mapper::get_barcode( $item, $fallback_barcode )`); now it's `$fallbacks[0]`, which is `''` on the harvest path. `maybe_merge_labels()` stamps whatever it's handed onto the merged record (`Order/Base.php:1163-1172`), so every path except single-label-plus-A6 produces a merged record with an empty barcode. `harvest_barcodes_or_fail()` then deletes the just-written labels and throws "PostNL returned a label without a barcode" (`Order/Base.php:988-1000`) — meaning with A4 selected, every label save on the harvest path fails with a misleading error, single-collo included.

`Order\Base::resolve_parent_barcode()` (`Order/Base.php:1017`) exists for exactly this case — falling back to the first record's barcode when the prefetched one is empty. Using it here (or falling back to the first response item's barcode) restores main's behavior.

src/Rest_API/V4/Label/Service.php
```diff
@@ -307,6 +309,7 @@ private function extract_fields( Shipping\Item_Info $item_info, array $mapped, a
 			'weight_gr'     => (int) ( $item_info->shipment['total_weight'] ?? 0 ),
 			'reference'     => (string) ( $item_info->shipment['order_number'] ?? '' ),
 			'barcode'       => (string) ( $post_data['main_barcode'] ?? '' ),
+			'barcodes'      => array_values( array_filter( (array) ( $post_data['barcodes'] ?? array() ), 'is_scalar' ) ),
```
tommyshellberg: Legacy caps the wire collo count at 10: the `num_labels` sanitizer clamps it (`Legacy/Shipping/Item_Info.php:732-740`) and the V1 client loops that clamped value. This count instead comes from `count( $post_data['barcodes'] )`, and `maybe_create_multi_barcodes()` reads the raw backend value with no cap (`Order/Base.php:919-921`) — the admin field has `min => 0` and no max. An admin who types 15 ships 10 colli on V1 and 15 on V4, and the eligibility signal saw the clamped 10 while the request carried 15. Cheapest fix: slice this array to the parsed `num_labels`, so the request agrees with the same clamp legacy applies.

src/Rest_API/V4/Label/Service.php
```diff
@@ -462,26 +465,66 @@ private function build_client(): PostnlClientInterface {
...
 			throw new \Exception(
 				esc_html__( 'Cannot create the label. No shipment was returned by PostNL.', 'postnl-for-woocommerce' )
 			);
 		}
 
-		$barcode         = Response_Mapper::get_barcode( $item, $fallback_barcode );
+		$parent_barcode = (string) ( $fallbacks[0] ?? '' );
+		$records        = array();
+
+		foreach ( $items as $index => $item ) {
```
tommyshellberg, line 513: This loop trusts the response count in both directions. If PostNL returns 2 items for 3 pre-issued barcodes, 2 labels are written and merged — while `Order\Base` persists all 3 prefetched barcodes into meta (`Order/Base.php:660-668`). The merchant sees 3 barcodes and a 2-label sheet, and the third barcode was never confirmed with PostNL. More items than barcodes is also accepted silently. Nothing compares `count( $items )` to `count( $fallbacks )`; this service has `$this->logger`, so a warning on mismatch would make both cases diagnosable.

src/Rest_API/V4/Label/Service.php
```diff
@@ -462,26 +465,66 @@ private function build_client(): PostnlClientInterface {
...
+		if ( empty( $records ) ) {
+			throw new \Exception(
+				esc_html__( 'Cannot create the label. Label content is missing', 'postnl-for-woocommerce' )
+			);
+		}
+
+		// The merge helper rebuilds a fresh record for a multi-collo shipment, so
+		// re-attach the parent collo's international partner refs (both empty for a
+		// domestic shipment) — the single-collo record already carries them.
+		return $this->finalize_label_records(
```
tommyshellberg, line 562: Only collo 1's partner refs survive an international multi-collo shipment. `item_label_records()` builds each collo's records with that collo's own partner barcode, but multi-collo always takes the merge rebuild path, which discards them, and this call re-attaches `$items[0]`'s values only. EU/ROW multi-collo is reachable — `num_labels` is a standard field for EU and ROW destinations (`Helper/Mapping.php:130-137`) — so colli 2..N's partner barcodes never reach order meta. Nothing reads label-record partner refs yet, so nothing breaks today, but it quietly undoes what #347's capture was for on multi-collo orders. If keeping them matters, they need a home that survives the merge (e.g. per-collo entries alongside `merged_files`); if dropping them is fine, the comment above this call should say so explicitly.

src/Rest_API/V4/Label/Request_Builder.php
```diff
@@ -112,27 +116,53 @@ public static function build( array $fields ): ShipmentDeliveryRequest {
...
+	 *
+	 * A multi-collo shipment sends one item per barcode in a single request; the
+	 * SDK derives itemCount from the item count. Each collo carries the same
+	 * shipment reference and weight, matching the legacy per-shipment payload.
+	 *
+	 * @param array $fields Builder input keyed as documented on build().
+	 * @return ShippingItem[]
+	 */
+	private static function items( array $fields ): array {
+		$barcodes = array_values( array_filter( (array) ( $fields['barcodes'] ?? array() ), 'is_scalar' ) );
```
tommyshellberg: This filter already ran: `extract_fields()` applies the identical `array_filter( ..., 'is_scalar' )` before the builder ever sees the array (`Service.php:312`), and no other production caller reaches `build()`. Deleting either copy leaves the whole suite green (mutation-tested), and producing a non-scalar entry requires hand-corrupted order meta. Either give this copy a test that feeds a non-scalar through `build()` directly, or drop one of the two — duplicate defense with no test is how the next person "simplifies" the wrong one.

tommyshellberg (changes requested):
**Terms used below, for anyone new to this code.** A "collo" is one box in a shipment (plural: colli). An order with "Number of labels: 3" ships as 3 boxes — 3 colli. Each box gets its own barcode. "V1" (or "legacy") is the PostNL API the plugin uses in production today. "V4" is the new API this migration adds. All V4 code sits behind a feature flag that is off by default.

Barcodes can be created two ways. Today the plugin creates barcodes first, then asks for labels ("pre-issued" barcodes). A planned future mode (task 14) does the reverse: the label call itself creates the barcodes, and the plugin reads them out of the response. I call that the "harvest path" below.

**Update: all six inline comments are fixed. The thread replies carry the commit hashes.** The text below is the original review, edited for clarity.

**Summary.** The multi-collo feature works correctly with pre-issued barcodes. The request sends one item per box. The saved order data matches what V1 produces. Label merging reuses the exact shared code V1 uses, so its behavior matches too — including the v5.7.2 fix for which slot on the label sheet printing starts at. The problems were on the harvest path, plus two smaller things:

1. On the harvest path, a 3-box order was sent to PostNL as 1 box. No error, no log. (Fixed.)
2. On the harvest path with A4 paper, every label save failed, with a wrong error message. (Fixed.)
3. V1 caps a shipment at 10 boxes. V4 didn't. Typing 15 shipped 10 boxes on V1 and 15 on V4. (Fixed.)
4. The new `store_labels()` code had no tests at all. (Fixed — it is now covered.)

**Mutation testing: 11 mutations, 7 killed on the reviewed code.** After the fix round, the full set was re-run: 13 of 13 killed.

| Mutation | Result |
|---|---|
| Eligibility: revert to `1 !== num_labels` | RED — `test_multi_collo_parcel_is_eligible` |
| Eligibility: delete the guard | RED — `test_ineligible_cases` "zero collo" |
| Builder: drop `is_scalar` filter | survived — filter since removed as duplicate |
| Builder: ignore `barcodes`, always single item | RED — `test_builds_multi_collo_items` |
| Builder: stop after first barcode | RED — `test_builds_multi_collo_items` |
| Builder: weight clamp to 0 | RED — 3 tests |
| store_labels: drop per-index fallback | survived then — now RED |
| store_labels: keep only first response item | survived then — now RED |
| store_labels: blank partner refs | survived then — now RED |
| Response_Mapper: return first item only | RED — `test_all_shipment_items_returns_all_collos` |
| extract_fields: drop the `barcodes` key | survived then — now RED |

The two guard mutations fail different tests in each direction. No new test fixture equals a fallback value. That trap appeared in the last four PRs and did not repeat here. The four survivors were all in `Service.php`, because `ServiceTest` had not been extended to the new code. All four are covered now.

**Checked and fine — no need to re-verify:**
- Weight: V1 repeats the full order weight on every box. V4 does the same.
- The SDK computes `itemCount` by counting the items. The count can never disagree with the list.
- Saved order data has the same shape as V1 output, plus an `api_version` tag.
- The `postnl_order_weight` filter still runs. It lives in shared code both APIs use. No setting or filter is newly lost by this PR.
- Insurance is declared once per shipment, matching V1 (V1 puts it only on the first box).
- The `< 1` guard in `Eligibility` can never fire in production. The form value is normalized to a whole number between 1 and 10 before the guard sees it. Harmless. Just know the "zero collo" test pins a state that cannot happen.

**Open question for PostNL, recorded as Q15 in the flip checklist.** V1 explicitly told the API "these 3 packages belong together, under this main barcode" (its Group fields). V4 has no such fields anywhere in the SDK. We send 3 items and assume PostNL links them into one shipment. Nobody has confirmed that. Nobody has confirmed Track & Trace shows them as one shipment. Nobody has confirmed response items come back in request order. All three need a sandbox test before this flag ever turns on.

**Rebase note.** This branch was rebased onto main after #347 merged. Your last commit and main's #347 fix solved the same problem: keys being dropped when labels merge. I kept main's version (`finalize_label_records()`) and wired your new code through it, because main's tests point at that method directly. Behavior is identical. Say the word if you want your inline version back.

Small description fix: the test count says 458. The suite on this branch is now 614.

src/Rest_API/V4/Label/Service.php
tommyshellberg: Fixed in 87aea67. `extract_fields()` now carries the schema-clamped `num_labels`, and `items()` builds that many barcode-less items when no barcodes were pre-issued — so a 3-collo order sends 3 items on the harvest path instead of silently collapsing to 1. The docblock records the one open assumption (N barcode-less items → N auto-issued labels is not covered by the SDK docs; if the API rejects it, it fails loudly, which is the point). Sandbox item is in the flip-checklist (Q15c). Mutation-verified both ways: reverting to a single item fails `test_builds_one_barcode_less_item_per_collo`; using num_labels even with barcodes present fails `test_builds_multi_collo_items` and `test_pre_issued_barcodes_win_over_num_labels`.

src/Rest_API/V4/Label/Service.php
tommyshellberg: Fixed in fba830c. `store_labels()` now resolves the merge key through the existing `Order\Base::resolve_parent_barcode()` — the pre-issued barcode still wins when present, and on the harvest path the first response-issued barcode is used, restoring main's behavior. This commit also adds the first `store_labels()`-level tests (a merge-spy subclass returning the exact Base.php:1165 rebuilt shape): reverting to raw `fallbacks[0]` fails `test_merge_is_keyed_on_the_response_barcode_when_none_was_preissued`.

src/Rest_API/V4/Label/Service.php
tommyshellberg: Fixed in c225d1c. `extract_fields()` slices the barcode list to the parsed (schema-clamped) `num_labels`, so V4 sends the same 10-collo maximum V1 ships. Typing 15 now behaves identically on both paths: 10 colli on the wire, 5 pre-issued barcodes wasted — the wasted prefetch is legacy-side and out of scope here. Mutation-verified: removing the slice, slicing to 1, and hardcoding 10 each fail `test_extract_fields_clamps_pre_issued_barcodes_to_the_collo_count` (the test runs 12→10 and 12→3 cases so a hardcoded cap can't ship green).

src/Rest_API/V4/Label/Service.php
tommyshellberg: Fixed in a350c9c. `store_labels()` now logs a warning (order number + expected + received counts) whenever the response item count differs from the pre-issued barcode count, in both directions, and still proceeds — partial labels beat none, the warning is for diagnosability. The harvest path (no pre-issued barcodes) correctly never warns. The two directions are pinned by different tests: deleting the warn call fails the fewer/more tests; dropping the non-empty filter fails `test_the_harvest_path_is_never_a_collo_count_mismatch`.

src/Rest_API/V4/Label/Service.php
tommyshellberg: Documented in a350c9c rather than changed: the comment above the finalize call now states plainly that each collo's own partner refs are captured per-collo but discarded by the merge rebuild, so only the parent collo's reach order meta. Persisting per-collo refs would change the meta shape, which felt like your call rather than a review fix — happy to do it in a follow-up if partner track-and-trace per collo is wanted. The items[0] wiring itself is now pinned by `test_the_merged_record_carries_the_parent_collos_partner_references` (distinct per-collo fixtures, so wiring collo 2's refs by mistake fails).

src/Rest_API/V4/Label/Request_Builder.php
tommyshellberg: Resolved in b351d9b by deleting this copy. `build()` has exactly one production caller (`Service::create()`, grep-verified repo-wide) and `extract_fields()` already normalizes the array on that path, so the entry point keeps the guard and the builder trusts its contract — which is now documented on the `@type array $barcodes` line. All 614 tests stay green after the deletion.

src/Rest_API/V4/Label/Service.php
tommyshellberg: Update: Thomas decided to keep the data. Done in 9f92433. The record keeps the flat parent keys unchanged and adds one new key, `partner_references`: one entry per box (its barcode, partner barcode and partner id), only present when at least one box has partner data. Nothing reads it yet — it is stored so a later per-box partner tracking feature has the numbers, and we can drop or reshape it once the sandbox shows what the V4 API actually returns per box (checklist Q15). Pinned both ways by different tests: removing the attach fails `test_every_collos_partner_references_are_kept_on_the_merged_record`; attaching unconditionally fails `test_a_shipment_with_no_partner_data_gains_no_partner_references_key`.

tommyshellberg (approved):
Approving. All six review findings are fixed and each fix is pinned by tests that fail in both directions (13 of 13 mutations killed on the final code). The partner tracking numbers question is settled: every box's numbers are now stored (9f92433, Thomas's call), and we can reshape that once the sandbox shows what the V4 API actually returns per box.

Reminder of what this PR does NOT need but the flag flip DOES (all recorded in the flip checklist): the sandbox tests for Q15 — does the V4 API group N items into one shipment, does Track & Trace show them as one, and do N barcode-less items produce N labels. None of that blocks merging, because the flag is off and no V4 key exists.

One step left before merge per our process: the staging check that the live V1 path is unchanged.

PR #349, opened 2026-07-25
22. V4 Returns service (return/generate)

sal4sup:
### All Submissions:

ClickUp: [22. V4 Returns service (return/generate)](https://app.clickup.com/t/868jqaqhh)

Adds a unified V4 Returns service backing `POST /shipment/delivery/v4/return/generate`. PostNL confirmed (2026-05-21) the three legacy return mechanisms collapse onto this one endpoint, differentiated by `printMethod` and receiver address. Stacked on `postnl-v4-shipping-label-domestic-parcel` (task 18) because it shares the SDK plumbing (`Exception_Converter`) introduced there.

- **`V4\Returns\Request_Builder`** (pure) — builds the `ReturnShipmentRequest`: `sender` = the consumer returning the item (with contact), `receiver` = the merchant return address, `labelSettings.printMethod` = `retailPrint`, `shipmentType` = `parcel`, one item (barcode auto-issued when none supplied). **No `returnOptions` block** — `returnPeriod` is omitted (V4 default is 35 days; PostNL do not want it differentiated) and there are no additional return services. Merchant `customerNumber`/`customerCode` are injected by the SDK client, so they're absent here.
- **`V4\Returns\Response_Mapper`** (pure) — captures the issued barcode (item → `returnBarcode` → fallback), the label document(s), and the international partner refs.
- **`V4\Returns\Service`** — implements `Return_Label_Service_Interface`. `create()` builds the request from the order + return-address settings, calls `returns()->generateReturn()`, and stores the result via the shared `Order\Base` helpers as `_postnl_order_metadata['labels']['return-label']`. Only the NL `retailPrint` flow is handled; a non-NL origin (consumerPrint is a BE follow-on) falls back to the legacy pipeline. `activate()` is unchanged — it still delegates to the legacy V1 `/parcels/v1/shipment/activatereturn` endpoint, which is not part of this migration.

**Reply-number vs return-to-home is encoded in the receiver address** (resolves open question §14): the freepost/Antwoordnummer receiver (Street `Antwoordnummer`, house number = reply number, freepost zip/city) vs the merchant home address, gated by `Settings::is_return_to_home_enabled()` — both produce valid requests. Both gates (validated New API Key + the return flow flag) default off, so merging changes nothing for merchants.

1. Run the PHP unit suite: `vendor/bin/phpunit -c phpunit-unit.xml.dist` — all pass (18 new Returns tests).
2. `composer check-php` — no violations on the new files.
3. Review the request DTO tests: `retailPrint` + parcel, `returnOptions`/`returnPeriod` omitted, no service flags, and both reply-number and return-to-home receiver addressing serialize to valid requests.
4. With a validated New API Key and the return flow flag on a NL store, generate a return label for an order and confirm the request hits `return/generate` with `printMethod: retailPrint` and the correct receiver address for the configured reply-number vs home setting; the returned barcode + label are stored as `labels['return-label']`.

### Other information:

### Changelog entry

Internal (no merchant-visible change): add a V4 SDK Returns service generating NL retailPrint return labels via `return/generate`, behind the default-off return flow flag.

### DIFF
--- a/src/Rest_API/Service_Factory.php
+++ b/src/Rest_API/Service_Factory.php
@@ -101,6 +101,17 @@ class Service_Factory {
 	private $label_v4_memo = null;
 
 	/**
+	 * Memoised self-built V4 returns service.
+	 *
+	 * Kept out of $v4_services for the same reason as $label_v4_memo: that array
+	 * means "a V4 service was explicitly injected for this flow", and giving it a
+	 * second meaning is how a later predicate reading it gets a wrong answer.
+	 *
+	 * @var V4_Returns_Service|null
+	 */
+	private $return_label_v4_memo = null;
+
+	/**
 	 * Service_Factory constructor.
 	 *
 	 * @param object|null $settings Plugin settings instance, or null when unavailable.
@@ -250,12 +261,25 @@ class Service_Factory {
 	 */
 	public function return_label_service(): Return_Label_Service_Interface {
 		if ( $this->should_use_v4( 'return_label' ) ) {
-			// A test double injected via inject_v4_service() wins; otherwise build the real V4 service,
-			// which handles the NL retailPrint return and falls back to the legacy pipeline for the rest.
-			if ( ! isset( $this->v4_services['return_label'] ) ) {
-				$this->v4_services['return_label'] = new V4_Returns_Service();
+			// A service injected via inject_v4_service() wins; otherwise build the real V4
+			// service, which handles the NL retailPrint return and falls back to the legacy
+			// pipeline for the rest.
+			if ( isset( $this->v4_services['return_label'] ) ) {
+				return $this->v4_services['return_label'];
 			}
-			return $this->v4_services['return_label'];
+			// Memoised apart from $v4_services for the same reason as label_service():
+			// that array means "deliberately injected", and barcode_from_label() reads it.
+			// Caching a self-built instance there would give the array two meanings, and
+			// the next predicate written against it would silently get the wrong answer.
+			if ( null === $this->return_label_v4_memo ) {
+				$logger                     = $this->v4_logger();
+				$this->return_label_v4_memo = new V4_Returns_Service(
+					new Client_Factory( $this->settings, $logger ),
+					(string) $this->settings->get_api_key_new(),
+					$logger
+				);
+			}
+			return $this->return_label_v4_memo;
 		}
 		if ( ! isset( $this->legacy_memos['return_label'] ) ) {
 			$this->legacy_memos['return_label'] = new Legacy_Return_Label_Service();
--- a/src/Rest_API/Service_Factory.php
+++ b/src/Rest_API/Service_Factory.php
@@ -27,6 +27,7 @@ use PostNLWooCommerce\Rest_API\Legacy\Smart_Returns_Service as Legacy_Smart_Retu
 use PostNLWooCommerce\Rest_API\SDK\Client_Factory;
 use PostNLWooCommerce\Rest_API\SDK\Logger_Adapter;
 use PostNLWooCommerce\Rest_API\V4\Label\Service as V4_Label_Service;
+use PostNLWooCommerce\Rest_API\V4\Returns\Service as V4_Returns_Service;
 use Psr\Log\LoggerInterface;
 
 if ( ! defined( 'ABSPATH' ) ) {
@@ -248,7 +249,12 @@ class Service_Factory {
 	 * @return Return_Label_Service_Interface
 	 */
 	public function return_label_service(): Return_Label_Service_Interface {
-		if ( $this->should_use_v4( 'return_label' ) && isset( $this->v4_services['return_label'] ) ) {
+		if ( $this->should_use_v4( 'return_label' ) ) {
+			// A test double injected via inject_v4_service() wins; otherwise build the real V4 service,
+			// which handles the NL retailPrint return and falls back to the legacy pipeline for the rest.
+			if ( ! isset( $this->v4_services['return_label'] ) ) {
+				$this->v4_services['return_label'] = new V4_Returns_Service();
+			}
 			return $this->v4_services['return_label'];
 		}
 		if ( ! isset( $this->legacy_memos['return_label'] ) ) {
--- a/src/Rest_API/V4/Returns/Request_Builder.php
+++ b/src/Rest_API/V4/Returns/Request_Builder.php
@@ -12,6 +12,7 @@ namespace PostNLWooCommerce\Rest_API\V4\Returns;
 use Postnl\Sdk\Enums\Payload\Country;
 use Postnl\Sdk\Enums\Payload\LabelOutputType;
 use Postnl\Sdk\Enums\Payload\LabelPrintMethod;
+use Postnl\Sdk\Enums\Payload\LabelResolution;
 use Postnl\Sdk\Enums\Payload\ShipmentType;
 use Postnl\Sdk\RequestData\V4\Address;
 use Postnl\Sdk\RequestData\V4\Contact;
@@ -68,7 +69,8 @@ class Request_Builder {
 	 *                                company, street, house_number, house_number_ext,
 	 *                                postcode, city, country.
 	 *     @type string $print_method LabelPrintMethod value, e.g. 'retailPrint'.
-	 *     @type array  $label        Label output: output_type (pdf|zpl|jpg|gif|png).
+	 *     @type array  $label        Label output: output_type (pdf|zpl|jpg|gif|png)
+	 *                                and resolution (200|300|600).
 	 *     @type string $barcode      Pre-issued return barcode to confirm; empty to let
 	 *                                the endpoint auto-issue one.
 	 *     @type string $reference    Merchant shipment reference (order number).
@@ -92,6 +94,7 @@ class Request_Builder {
 
 		$label_settings = new LabelSettings(
 			outputType: self::output_type( (string) ( $label_fields['output_type'] ?? 'pdf' ) ),
+			resolution: self::resolution( (int) ( $label_fields['resolution'] ?? 200 ) ),
 			printMethod: self::print_method( (string) ( $fields['print_method'] ?? 'retailPrint' ) )
 		);
 
@@ -178,6 +181,20 @@ class Request_Builder {
 	 */
 	private static function output_type( string $output_type ): LabelOutputType {
 		return LabelOutputType::tryFrom( strtolower( $output_type ) ) ?? LabelOutputType::PDF;
+	}
+
+	/**
+	 * Resolve a resolution integer into the SDK LabelResolution enum.
+	 *
+	 * The merchant's DPI setting offers 200, 300 and 600 and defaults to 600. The
+	 * SDK's own default is 200, so leaving this unset silently downgraded every
+	 * return label.
+	 *
+	 * @param int $resolution Print resolution in DPI.
+	 * @return LabelResolution
+	 */
+	private static function resolution( int $resolution ): LabelResolution {
+		return LabelResolution::tryFrom( $resolution ) ?? LabelResolution::DPI_200;
 	}
 
 	/**
--- a/src/Rest_API/V4/Returns/Request_Builder.php
+++ b/src/Rest_API/V4/Returns/Request_Builder.php
@@ -1,0 +1,192 @@
+<?php
+/**
+ * Class Rest_API\V4\Returns\Request_Builder file.
+ *
+ * @package PostNLWooCommerce\Rest_API\V4\Returns
+ */
+
+declare( strict_types = 1 );
+
+namespace PostNLWooCommerce\Rest_API\V4\Returns;
+
+use Postnl\Sdk\Enums\Payload\Country;
+use Postnl\Sdk\Enums\Payload\LabelOutputType;
+use Postnl\Sdk\Enums\Payload\LabelPrintMethod;
+use Postnl\Sdk\Enums\Payload\ShipmentType;
+use Postnl\Sdk\RequestData\V4\Address;
+use Postnl\Sdk\RequestData\V4\Contact;
+use Postnl\Sdk\RequestData\V4\CustomerReferences;
+use Postnl\Sdk\RequestData\V4\Dimensions;
+use Postnl\Sdk\RequestData\V4\LabelSettings;
+use Postnl\Sdk\RequestData\V4\ShipmentParty;
+use Postnl\Sdk\ResponseData\V4\ShippingItem;
+use Postnl\Sdk\Service\ReturnShipment\V4\Request\ReturnShipmentRequest;
+
+if ( ! defined( 'ABSPATH' ) ) {
+	exit;
+}
+
+/**
+ * Class Request_Builder
+ *
+ * Pure translator from a flat, already-parsed field array into a V4
+ * ReturnShipmentRequest DTO for the /shipment/delivery/v4/return/generate
+ * endpoint. It performs no WooCommerce or settings access, so the DTO shape can
+ * be asserted in isolation.
+ *
+ * Scope: the retail-print (retailPrint) NL flow. PostNL confirmed (2026-05-21)
+ * the three legacy return mechanisms collapse onto this one endpoint,
+ * differentiated by printMethod and the receiver address:
+ *
+ * - printMethod is retailPrint for NL-origin returns (consumerPrint is a
+ *   BE-origin follow-on, left unimplemented).
+ * - The sender is the consumer returning the item; the receiver is the merchant,
+ *   whose address encodes reply-number (Antwoordnummer) vs return-to-home.
+ * - returnPeriod is omitted (the V4 default is 35 days and PostNL do not want it
+ *   differentiated) and there are no additional services for return products, so
+ *   no ReturnOptions block is emitted.
+ *
+ * The merchant's customerNumber/customerCode are injected into the receiver by
+ * the SDK client (ClientBuilder::withCustomerCredentials), so they are
+ * deliberately absent here.
+ *
+ * @since   5.9.10
+ * @package PostNLWooCommerce\Rest_API\V4\Returns
+ */
+class Request_Builder {
+
+	/**
+	 * Build the ReturnShipmentRequest from parsed order/settings fields.
+	 *
+	 * @param array $fields {
+	 *     Flat, pre-parsed values.
+	 *
+	 *     @type array  $sender       Consumer returning the item: company, first_name,
+	 *                                last_name, street, house_number, house_number_ext,
+	 *                                postcode, city, country, email, phone.
+	 *     @type array  $receiver     Merchant return address (reply-number or home):
+	 *                                company, street, house_number, house_number_ext,
+	 *                                postcode, city, country.
+	 *     @type string $print_method LabelPrintMethod value, e.g. 'retailPrint'.
+	 *     @type array  $label        Label output: output_type (pdf|zpl|jpg|gif|png).
+	 *     @type string $barcode      Pre-issued return barcode to confirm; empty to let
+	 *                                the endpoint auto-issue one.
+	 *     @type string $reference    Merchant shipment reference (order number).
+	 *     @type int    $weight_gr    Total shipment weight in grams.
+	 * }
+	 * @return ReturnShipmentRequest
+	 */
+	public static function build( array $fields ): ReturnShipmentRequest {
+		$sender_fields   = $fields['sender'] ?? array();
+		$receiver_fields = $fields['receiver'] ?? array();
+		$label_fields    = $fields['label'] ?? array();
+
+		$sender = ShipmentParty::asReturnSender(
+			address: self::address( $sender_fields ),
+			contact: self::contact( $sender_fields )
+		);
+
+		$receiver = ShipmentParty::asReturnReceiver(
+			address: self::address( $receiver_fields )
+		);
+
+		$label_settings = new LabelSettings(
+			outputType: self::output_type( (string) ( $label_fields['output_type'] ?? 'pdf' ) ),
+			printMethod: self::print_method( (string) ( $fields['print_method'] ?? 'retailPrint' ) )
+		);
+
+		$item = new ShippingItem(
+			barcode: self::maybe_null( (string) ( $fields['barcode'] ?? '' ) ),
+			customerReferences: new CustomerReferences(
+				shipmentReference: self::maybe_null( (string) ( $fields['reference'] ?? '' ) )
+			),
+			dimensions: new Dimensions(
+				weightGr: max( 1, (int) ( $fields['weight_gr'] ?? 0 ) )
+			)
+		);
+
+		return new ReturnShipmentRequest(
+			sender: $sender,
+			receiver: $receiver,
+			labelSettings: $label_settings,
+			shipmentType: ShipmentType::Parcel,
+			items: array( $item )
+		);
+	}
+
+	/**
+	 * Translate an address field array into a V4 Address DTO.
+	 *
+	 * @param array $fields Address fields keyed as documented on build().
+	 * @return Address
+	 */
+	private static function address( array $fields ): Address {
+		return new Address(
+			countryIso: self::country( (string) ( $fields['country'] ?? '' ) ),
+			houseNumber: self::maybe_null( (string) ( $fields['house_number'] ?? '' ) ),
+			postalCode: self::maybe_null( (string) ( $fields['postcode'] ?? '' ) ),
+			companyName: self::maybe_null( (string) ( $fields['company'] ?? '' ) ),
+			street: self::maybe_null( (string) ( $fields['street'] ?? '' ) ),
+			houseNumberAddition: self::maybe_null( (string) ( $fields['house_number_ext'] ?? '' ) ),
+			city: self::maybe_null( (string) ( $fields['city'] ?? '' ) )
+		);
+	}
+
+	/**
+	 * Translate consumer contact fields into a V4 Contact DTO.
+	 *
+	 * @param array $fields Sender fields keyed as documented on build().
+	 * @return Contact
+	 */
+	private static function contact( array $fields ): Contact {
+		return new Contact(
+			email: self::maybe_null( (string) ( $fields['email'] ?? '' ) ),
+			firstName: self::maybe_null( (string) ( $fields['first_name'] ?? '' ) ),
+			lastName: self::maybe_null( (string) ( $fields['last_name'] ?? '' ) ),
+			mobileNumber: self::maybe_null( (string) ( $fields['phone'] ?? '' ) ),
+			companyName: self::maybe_null( (string) ( $fields['company'] ?? '' ) )
+		);
+	}
+
+	/**
+	 * Resolve a country code into the SDK Country enum, defaulting to NL.
+	 *
+	 * @param string $code Two-letter ISO country code.
+	 * @return Country
+	 */
+	private static function country( string $code ): Country {
+		return Country::tryFrom( strtoupper( $code ) ) ?? Country::NL;
+	}
+
+	/**
+	 * Resolve a print-method string into the SDK LabelPrintMethod enum.
+	 *
+	 * Defaults to retailPrint — the only method valid for the NL-origin flow.
+	 *
+	 * @param string $print_method LabelPrintMethod value, e.g. 'retailPrint'.
+	 * @return LabelPrintMethod
+	 */
+	private static function print_method( string $print_method ): LabelPrintMethod {
+		return LabelPrintMethod::tryFrom( $print_method ) ?? LabelPrintMethod::Retail;
+	}
+
+	/**
+	 * Resolve a label output-type string into the SDK LabelOutputType enum.
+	 *
+	 * @param string $output_type One of pdf|zpl|jpg|gif|png.
+	 * @return LabelOutputType
+	 */
+	private static function output_type( string $output_type ): LabelOutputType {
+		return LabelOutputType::tryFrom( strtolower( $output_type ) ) ?? LabelOutputType::PDF;
+	}
+
+	/**
+	 * Return null for an empty string so the DTO omits the field entirely.
+	 *
+	 * @param string $value Candidate value.
+	 * @return string|null
+	 */
+	private static function maybe_null( string $value ): ?string {
+		return '' === $value ? null : $value;
+	}
+}
--- a/src/Rest_API/V4/Returns/Response_Mapper.php
+++ b/src/Rest_API/V4/Returns/Response_Mapper.php
@@ -1,0 +1,148 @@
+<?php
+/**
+ * Class Rest_API\V4\Returns\Response_Mapper file.
+ *
+ * @package PostNLWooCommerce\Rest_API\V4\Returns
+ */
+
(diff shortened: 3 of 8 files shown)


src/Rest_API/V4/Returns/Service.php
```diff
@@ -0,0 +1,305 @@
...
+	 * Decide whether an order's return is the NL retailPrint flow this service handles.
+	 *
+	 * Only NL-origin returns use retailPrint; a BE origin (consumerPrint) is not
+	 * yet implemented and falls back to legacy.
+	 *
+	 * @param Item_Info $item_info Parsed legacy return item info.
+	 * @return bool
+	 */
+	private function is_eligible( Item_Info $item_info ): bool {
+		return 'NL' === (string) ( $item_info->shipper['country'] ?? '' );
```
tommyshellberg: This reads the store's country, not the customer's. So it does not do what the docblock above it says.

`$item_info->shipper` is the store address. `Legacy/Shipping/Item_Info.php:118` builds it from `$this->api_args['store_address']`. The plugin already refuses to run unless the store is in NL or BE, so on any Dutch shop this check is always true.

But a return is sent from wherever the customer is. The SDK spells that out on the `LabelPrintMethod` enum:

> Sending the return from BE means only "consumer" is available. Sending from NL means only "retail" is available.

And a Dutch shop can offer return labels to Belgian customers. `Mapping::option_available_list()` line 129 lists `create_return_label` under `NL => BE`.

So: Dutch shop, Belgian customer, merchant ticks "Create Return Label". This returns true, `extract_fields()` hardcodes `retailPrint`, and the request goes out with a print method the SDK says is not available for a parcel handed in at a Belgian counter. The docblock promises a non-NL origin falls back to legacy. It never will, because the check is looking at the wrong address.

One-line fix — gate on the customer's country instead:

```php
return 'NL' === (string) ( $item_info->receiver['country'] ?? '' );
```

A test with a Belgian customer and a Dutch store would pin it.

src/Rest_API/V4/Returns/Service.php
```diff
@@ -0,0 +1,305 @@
...
+	 * @var Client_Factory|null
+	 */
+	private $client_factory;
+
+	/**
+	 * Service constructor.
+	 *
+	 * @param Client_Factory|null $client_factory Optional SDK client factory.
+	 */
+	public function __construct( ?Client_Factory $client_factory = null ) {
```
tommyshellberg: This constructor takes an optional factory and works out everything else by itself. The label service next door does the opposite, and that difference was the whole point of three earlier fix rounds (#337, #339, #343).

`V4\Label\Service::__construct` on main requires all three things it needs:

```php
public function __construct(
    Client_Factory $client_factory,
    #[\SensitiveParameter]
    string $v4_key,
    LoggerInterface $logger
) {
```

This one takes none of them as required. It builds its own `Client_Factory`, digs the key out of settings in `build_client()` behind a `method_exists()` check, and never takes a logger at all. `Service_Factory` then calls `new V4_Returns_Service()` with no arguments, so there is nowhere for the wiring to go.

Why it matters: forgetting to pass a logger should break at the call site while you are writing it. Right now forgetting it produces silent code that compiles and passes tests. Making all three required turns that into a `TypeError` during development.

The `method_exists( $this->settings, 'get_api_key_new' )` guard in `build_client()` goes away with this change, which is a bonus — the fallback to `''` is a landmine. `Auth::apiKey('')` throws in the SDK (`ApiKeyAuth.php:23`), and `Client_Factory::build()` does not guard against it. It cannot fire today because `has_v4_key()` checks the same method name first, but it should not exist at all.

src/Rest_API/V4/Returns/Service.php
```diff
@@ -0,0 +1,305 @@
...
+			|| ! $this->is_eligible( $item_info ) ) {
+			return $this->maybe_create_return_label_pipeline( $post_data );
+		}
+
+		$fields  = $this->extract_fields( $item_info, $post_data );
+		$request = Request_Builder::build( $fields );
+
+		try {
+			$response = $this->build_client()->returns()->generateReturn( $request );
+		} catch ( \Throwable $exception ) {
```
tommyshellberg, line 170: Nothing on this path writes a log line. `grep -c logger src/Rest_API/V4/Returns/*.php` returns 0.

That is bad here specifically, because of what the merchant is told. `Exception_Converter::convert()` sends anything it does not recognise to this message:

> An unexpected error occurred while contacting PostNL. Check the PostNL logs for details.

The logs are guaranteed empty. The message points the merchant at a file where nothing was written.

Three gaps stack up to that. The service has no logger. `build_client()` calls `new Client_Factory( $this->settings )` with no second argument, so the SDK installs a null logger and no merchant setting can switch it on. And this catch block converts and rethrows without logging.

Compare what this sits alongside. `V4\Label\Service` logs the order reference, the destination, the merchant-facing message, and the original exception class and message. Legacy `Rest_API\Base::send_request()` logs the URL, the request arguments, the response headers, the response body, and any `WP_Error`. This would be the only PostNL call in the plugin that can fail leaving no trace.

Fix along with the constructor comment: take a required `LoggerInterface`, pass it into `Client_Factory`, and log here the way the label service does.

src/Rest_API/V4/Returns/Service.php
```diff
@@ -0,0 +1,305 @@
...
+	 * untouched legacy pipeline.
+	 *
+	 * @param array $post_data Context needed to build and send the return request.
+	 *
+	 * @return array Normalized label record keyed by 'return-label'.
+	 *
+	 * @throws \Exception If the SDK request fails (converted to the legacy error shape).
+	 */
+	public function create( array $post_data ): array {
+		$item_info = new Item_Info( $post_data );
```
tommyshellberg, line 141: `Item_Info` is built here, before the check on line 93 that asks whether the merchant actually wanted a return label. Legacy does it the other way round.

`Order/Base.php:1451-1458` checks `create_return_label` first and returns an empty array. It only builds `Item_Info` after that check passes.

The order matters because this constructor throws. `Return_Label\Item_Info` extends `Shipping\Item_Info`, which lists `main_barcode` as required with no default and throws "Barcode is empty!" when it is missing (`Legacy/Shipping/Item_Info.php:434`). On the V4 label path `main_barcode` is never set — `Order/Base.php:629-636` takes the branch that harvests the barcode out of the label response instead of pre-ordering one.

So with the label flag and the return flag both on, an order where the merchant left the box unticked would throw "Barcode is empty!" — an error about barcodes, on a save where no return label was wanted. Legacy returns quietly.

This is the same shape as the finding on #343. Moving the guard above line 89 fixes it.

Related: deleting this whole guard leaves the suite green. Nothing tests it.

src/Rest_API/V4/Returns/Request_Builder.php
```diff
@@ -0,0 +1,192 @@
...
+		$sender = ShipmentParty::asReturnSender(
+			address: self::address( $sender_fields ),
+			contact: self::contact( $sender_fields )
+		);
+
+		$receiver = ShipmentParty::asReturnReceiver(
+			address: self::address( $receiver_fields )
+		);
+
+		$label_settings = new LabelSettings(
```
tommyshellberg, line 95: The merchant's print resolution is dropped here.

`Settings::get_printer_type()` builds strings like `GraphicFile|GIF 600 dpi` from the `printer_type_resolution` setting, which offers 600, 300 and 200 and defaults to **600**. `Service::resolve_output_type()` pulls the format out of that string and throws the number away.

The SDK has a field for it. `LabelSettings` takes a `resolution` parameter, and its own docblock says the default is **200 DPI**.

So a merchant set to 600 DPI gets 200 DPI return labels, and nothing tells them. On a label printer that is a visible drop in print quality and can make the barcode harder to scan.

The field exists, so this is a real loss rather than something the SDK cannot express. Suggest passing the resolution through: read it in `extract_fields()` alongside the output type and map it onto `LabelResolution`.

src/Rest_API/V4/Returns/Service.php
```diff
@@ -0,0 +1,305 @@
...
+	 * Resolve the merchant's combined printer-type string to a discrete output type.
+	 *
+	 * PostNL advises PNG/JPG for retailPrint, but the merchant's configured format
+	 * is respected; PDF is the fallback.
+	 *
+	 * @param string $printer_type Legacy combined printer-type string.
+	 * @return string
+	 */
+	private function resolve_output_type( string $printer_type ): string {
+		foreach ( array( 'zpl', 'jpg', 'gif', 'png', 'pdf' ) as $candidate ) {
```
tommyshellberg: Question rather than a defect, but it needs an answer before the flag is flipped.

This list is checked in order, so a Zebra printer wins first. `Settings::get_printer_type()` can only ever return one of four strings, and they map like this:

| Printer Type setting | string produced | output type sent |
|---|---|---|
| PDF (the default) | `GraphicFile\|PDF` | `pdf` |
| JPG | `GraphicFile\|JPG 600 dpi` | `jpg` |
| GIF | `GraphicFile\|GIF 600 dpi` | `gif` |
| ZPL | `Zebra\|Generic ZPL II 600 dpi` | `zpl` |

`png` is in the candidate list but unreachable — no setting produces it.

Two things point at a problem. The SDK comments its own enum case: `case ZPL = 'zpl'; // ZPL: not applicable to return endpoints`. And PostNL's guidance in the task says retailPrint output should be **PNG or JPG**.

So of the four settings a merchant can choose, one (JPG) matches PostNL's advice, and the default (PDF) does not. Whether the endpoint rejects PDF and ZPL outright, or accepts them and returns something unusable, I could not check — that needs a sandbox key.

Worth asking PostNL, and worth a sandbox item before the flip. If PDF is rejected, the default setting breaks every return label, which is the worst case to discover in production. If the answer is "PNG or JPG only", the fix is probably to map anything else to JPG for this flow rather than passing the merchant's choice straight through.

src/Rest_API/Service_Factory.php
```diff
@@ -248,7 +249,12 @@ public function letterbox_service(): Label_Service_Interface {
 	 * @return Return_Label_Service_Interface
 	 */
 	public function return_label_service(): Return_Label_Service_Interface {
-		if ( $this->should_use_v4( 'return_label' ) && isset( $this->v4_services['return_label'] ) ) {
+		if ( $this->should_use_v4( 'return_label' ) ) {
+			// A test double injected via inject_v4_service() wins; otherwise build the real V4 service,
+			// which handles the NL retailPrint return and falls back to the legacy pipeline for the rest.
+			if ( ! isset( $this->v4_services['return_label'] ) ) {
+				$this->v4_services['return_label'] = new V4_Returns_Service();
```
tommyshellberg: This puts a self-built service into `$v4_services`, which the comment about forty lines above says must not happen.

I checked whether it breaks anything today, and it does not. There are nine reads of `$v4_services` in this class and each one looks at a single fixed key. `barcode_from_label()` on line 169 reads `['label']` only, so writing `['return_label']` changes no decision anywhere. `inject_v4_service()` is a plain assignment, so a self-built instance cannot block a later deliberate injection either. No bug.

What it does break is the rule. `label_service()` keeps its self-built instance in a separate `$label_v4_memo` field, and the comment there explains why: `$v4_services` means "someone deliberately injected this", and `barcode_from_label()` depends on that meaning. Two methods in the same file now follow two different rules.

The cost lands later. The next person who adds a check like "has any V4 service been injected" gets the wrong answer, and it fails quietly — on #343 that exact shape crashed bulk label runs from the second order onward.

Four-line fix: add a private `$return_label_v4_memo` and leave `$v4_services` alone. A test that `inject_v4_service('return_label', ...)` still wins after `return_label_service()` has run once would pin it. There is already a matching pair of tests for the label flow in `Service_FactoryTest.php`.

tests/php/unit/Rest_API/V4/Returns/Request_BuilderTest.php
```diff
@@ -0,0 +1,204 @@
...
+			'receiver'     => array(
+				'company'          => 'My Shop',
+				'street'           => 'Siriusdreef',
+				'house_number'     => '42',
+				'house_number_ext' => '',
+				'postcode'         => '2132WT',
+				'city'             => 'Hoofddorp',
+				'country'          => 'NL',
+			),
+			'print_method' => 'retailPrint',
```
tommyshellberg, line 53: Two of the three fallback resolvers are not actually tested, because the fixtures sit on the fallback value.

I mutation-tested this. Replacing the whole body of `print_method()` with `return LabelPrintMethod::Retail;` leaves the suite green. So does replacing `country()` with `return Country::NL;`. So does deleting `strtoupper` and deleting `strtolower`.

Why: this fixture uses `'retailPrint'`, which is also the fallback, and `test_print_method_defaults_to_retail` feeds `'nonsense'` and also expects Retail. Both paths agree, so nothing proves the value is read from the input. Same for country — every address uses `'NL'`, which is the fallback, and is already uppercase so `strtoupper` never does any work.

Credit where it is due: `output_type()` is the one you got right. The fixture is `'jpg'` and the negative test uses `'bmp'`, so short-circuiting that resolver **does** fail. That is the pattern the other two need.

Three small changes fix all four survivors:

1. Add a test setting `print_method` to `'consumerPrint'` and asserting it comes through as `consumerPrint`. Keep the `'nonsense'` test — it is the pair that makes the fallback real.
2. Add a test with a lowercase non-NL sender country, `'be'`, asserting `countryIso` comes out `'BE'`. That one kills both the country survivors at once.
3. Change line 58 to `'output_type' => 'JPG'` and keep asserting `'jpg'`. That exercises `strtolower` without weakening anything.

This is the fifth PR in this queue with a fixture sitting on a fallback, so it is worth a glance on the next one too.

src/Rest_API/V4/Returns/Service.php
```diff
@@ -0,0 +1,305 @@
...
+	 * Routes the NL retailPrint return to the V4 SDK; a non-NL origin runs the
+	 * untouched legacy pipeline.
+	 *
+	 * @param array $post_data Context needed to build and send the return request.
+	 *
+	 * @return array Normalized label record keyed by 'return-label'.
+	 *
+	 * @throws \Exception If the SDK request fails (converted to the legacy error shape).
+	 */
+	public function create( array $post_data ): array {
```
tommyshellberg, line 129: This class has no tests. Nothing in `tests/` references it.

That is 305 lines holding the routing decision, the field flattening, the printer-type resolution, the file writing and the merge handling. Eleven mutations survive in here. All of these leave the suite green:

- flip the eligibility country check from NL to BE, or make it always true
- blank the barcode in `extract_fields()`
- send `consumerPrint` instead of `retailPrint`
- send weight 0 for every order
- force every output type to `pdf`
- delete either `continue` in `store_labels()`
- delete the `api_version = 'v4'` re-tag after merging
- delete the `create_return_label` guard

The constructor already accepts an injected `Client_Factory`, so most of this is reachable without WooCommerce. Six tests would cover the important half:

1. `create()` with `create_return_label` not `'yes'` falls back to the legacy pipeline and never calls the SDK.
2. A Belgian customer falls back to legacy; a Dutch customer calls `generateReturn`. (This is the check in the comment above — write it against the customer's country, not the store's.)
3. The request handed to `generateReturn` carries the return barcode from `$post_data`, `printMethod: retailPrint`, and the order weight.
4. A printer type containing `zpl` resolves to `'zpl'`, not `'pdf'`.
5. `store_labels()` throws "Label content is missing" when the only label decodes to empty.
6. Every record coming out of `store_labels()` carries `api_version => 'v4'` after the merge.

One more worth adding on the mapper: `decode_content()` currently passes strict base64 decoding, but both fixtures are valid base64, so removing the strict flag survives. A fixture like `'not!!valid!!base64'` asserting `''` would pin it. Without the strict flag a corrupt payload decodes to garbage and gets written to disk as a broken label file instead of being skipped.

src/Rest_API/V4/Returns/Response_Mapper.php
```diff
@@ -0,0 +1,148 @@
...
+		return $fallback;
+	}
+
+	/**
+	 * Return the international partner barcode issued for a return item.
+	 *
+	 * @param ReturnShippingItem $item Return item from the response.
+	 * @return string
+	 */
+	public static function get_partner_barcode( ReturnShippingItem $item ): string {
```
tommyshellberg: These two methods are written and tested, but nothing ever calls them. `grep -n partner src/Rest_API/V4/Returns/Service.php` comes back empty.

They also cannot ever be needed as the code stands. Returns here are Dutch only, and a partner barcode is an international thing — it identifies a delivery partner in another country.

The passing test makes this look wired up. It is not. The description lists partner references as something the mapper captures, which reads as a feature rather than dead code.

Either delete both methods and their test, or say in the docblock that they are there for the international return flow that has not been built yet. I would lean toward deleting — `V4\Label\Response_Mapper` already has its own copies, and those are called.

Small related note: `Service::store_labels()` re-stamps `api_version` after the merge, which is exactly right, because `maybe_merge_labels()` rebuilds the record and keeps only `type`, `barcode`, `created_at`, `filepath` and `merged_files`. Any key added in `to_label_record()` that is not re-applied afterwards would vanish on the merge path. Worth remembering if partner references ever do get stored.

tommyshellberg (changes requested):
I rebased this branch onto `main` before reading it, so the diff is now just this branch's own work: 6 files, +1006. Base is `main`. CI is green on both suites, PHPCS is clean on all four changed source files, and the unit suite is 634 green.

A few terms, since this area has its own vocabulary: **retailPrint** is the print method where the customer takes the parcel to a PostNL counter and staff print the label there. **Antwoordnummer** (reply number) is a Dutch freepost address — the merchant pays postage, and the address is a number instead of a street. **DPI** is print resolution.

## One thing to know before anyone tests this

`Service::create()` has no live caller. `Order\Base::maybe_create_return_label()` is called from exactly one place, `Order/Base.php:655`, and that line sits inside a `/* Temporarily commented. */` block running from line 651 to 658. The legacy return-create path is dead for the same reason — nothing else builds `Return_Label\Client`.

Return labels today come from the outbound label call instead, which carries a `ReturnBarcode` field.

Two consequences:

1. Step 4 of the test plan in the description cannot be performed. There is no code path that reaches `return/generate`, whatever the flags are set to. Worth correcting in the description so nobody loses an afternoon to it.
2. Whoever flips the flag later has to restore that commented block too. Otherwise the flag goes on and nothing happens at all — no error, no request. That belongs in the flip checklist.

This also caps how bad anything below can be. Every finding sits behind two gates that default off **and** a commented-out call site. Nothing here can reach a merchant today.

## Hard gates — all pass

- `composer.json` is untouched. The SDK is still in `require-dev`, pinned to `dev-main#5a064dd`, with the `php >= 7.4` floor still in `require`.
- Routing needs `has_v4_key()` (the validated new key, not `get_effective_api_key()`) **and** the `postnl_sdk_enable_return_label` filter. Both default off.
- Order meta shape is unchanged. `to_label_record()` writes `type`, `barcode`, `created_at`, `filepath`, `api_version`. `maybe_merge_labels()` rebuilds the record and drops `api_version`, but `store_labels()` re-stamps it afterwards, so nothing is lost. That is the right way round.
- No public filter or action signatures changed.
- The legacy branch of `Service_Factory::return_label_service()` is untouched.

## Mutation testing: 18 of 39 caught

What is genuinely solid: the parcel shipment type, sender and receiver ordering, the address contents for both the reply-number and home cases, the item barcode and shipment reference, the empty-string-to-null rule, the one-gram weight clamp, and the barcode preference order in the mapper. Adding a `returnPeriod` block to the request **does** fail `test_return_options_omitted`, so the main thing this task had to get right is properly pinned.

The survivors are covered in the comments below.

Two mutations I am **not** asking for tests on, because they cannot change behaviour:

- Dropping the `isEmpty()` check in `first_return_item()`. The SDK's `AbstractBaseCollection::first()` is `return $this->items[0] ?? null;` and the constructor stores `array_values($items)`, so both forms return the same thing for every input. The check is redundant, not untested.
- Changing the `?? 'NL'` receiver-country default in `extract_fields()`. `is_eligible()` runs first and already requires that value to be set, so the default cannot fire. Dead defensive code — worth deleting rather than testing.

## Checked and fine, so nobody re-treads these

- **Every SDK name this PR uses exists** at the pinned commit: `ShipmentParty::asReturnSender`, `asReturnReceiver`, `ReturnShipmentRequest`, `returns()->generateReturn()`. `ReturnShipmentRequest` really does type its items as `ResponseData\V4\ShippingItem` — the SDK reuses the response class on the request side, so that import is correct and not a mistake.
- **`returnOptions` really is omitted**, which is what PostNL asked for on 2026-05-21. The constructor argument is skipped entirely, and the test checks both the missing key and the serialized JSON.
- **The sender and receiver swap is right.** A return travels customer to merchant, so the outbound recipient becomes the return sender. The SDK docblocks confirm the direction.
- **No filters or actions are lost.** The legacy return path fires none at all — `grep` for `apply_filters` and `do_action` across `Legacy/Return_Label*` and `Legacy/Smart_Returns*` comes back empty. So the usual sweep has nothing to find here.
- **`new Legacy_Return_Label_Service()` inside `activate()` is harmless.** That class extends `Order\Base`, whose constructor calls `init_hooks()`, so it looked like it might register WordPress hooks on every call. Its `init_hooks()` is an empty no-op.
- **`'return-label'` is in `Mapping::label_type_list()`** for NL to NL and NL to BE, so the merge step does not filter these records away on the normal path.
- **All five acceptance criteria from the task have tests**: the retailPrint DTO, `returnPeriod` omitted, no service flags, and both reply-number and return-to-home addressing.

## What I could not check

Whether PostNL accepts `outputType: pdf` or `outputType: zpl` on the return endpoint. That needs a sandbox key, which does not exist yet. See the comment on `resolve_output_type()`.

src/Rest_API/V4/Returns/Service.php
tommyshellberg: Fixed in `c0a41c1`. `is_eligible()` now reads `$item_info->receiver['country']`, the customer's country.

Two tests pin it, and they fail on different mutations. Changing it back to the store address fails `test_belgian_customer_is_not_eligible`; making it always return true fails the same test. The fixture keeps the store Dutch and asserts that, so only the customer's country can decide the result.

The docblock now records why, including the SDK quote.

One thing this also closed by accident: restricting to Dutch customers keeps the merge step safe. The stored record is typed `return-label`, and `Mapping::label_type_list()` only lists that type for NL to NL and NL to BE. An EU or ROW destination would have left `maybe_merge_labels()` with no files to merge and hit an undefined index. That is now unreachable.

src/Rest_API/V4/Returns/Service.php
tommyshellberg: Fixed in `c0a41c1`. The constructor now requires all three, matching `V4\Label\Service`:

```php
public function __construct(
    Client_Factory $client_factory,
    #[\SensitiveParameter]
    string $v4_key,
    LoggerInterface $logger
) {
```

The `method_exists( get_api_key_new )` guard is gone with it, so the empty-key fallback no longer exists. `Service_Factory` supplies all three.

Pinned by `test_client_authenticates_with_the_injected_key`, which asserts the injected key reaches the `apiKey` header on the wire. Swapping `$this->v4_key` for any other value fails it.

src/Rest_API/V4/Returns/Service.php
tommyshellberg: Fixed in `c0a41c1`. The transport moved into a protected `generate_return()`, mirroring `V4\Label\Service::confirm_label()`, and its catch block logs the order number, the merchant-facing message, and the original SDK exception class and message.

The logger is a required constructor parameter and is passed into `Client_Factory`, so the SDK gets a real logger instead of a null one.

Mutated both ways, caught by different tests. Deleting the log call fails `test_failed_return_call_is_logged_with_the_original_cause`. Making it also log on success fails `test_successful_return_call_is_not_logged`. A logger that never fires and one that always fires are both caught.

src/Rest_API/V4/Returns/Service.php
tommyshellberg: Fixed in `c0a41c1`. The checkbox is checked first; `Item_Info` is built only after it passes.

Pinned by `test_create_without_the_checkbox_never_builds_item_info`. The post data in that test deliberately has no `main_barcode`, so putting the guard back after the constructor fails it — along with `test_create_falls_back_when_the_flag_is_absent`.

src/Rest_API/V4/Returns/Request_Builder.php
tommyshellberg: Fixed in `c0a41c1`, and it turned out there was already a shared helper for this.

`V4\Label\Request_Builder::printer_type_to_label_settings()` splits the combined setting into both an output type and a resolution, and the label path has been using it all along. The returns service had reimplemented only the output-type half. It now calls the shared helper, and the duplicated loop is deleted.

`Request_Builder` passes the value through as `resolution`.

Four tests pin it. Removing `resolution` from `LabelSettings` fails `test_resolution_is_sent` and `test_unknown_resolution_falls_back_to_200`. Dropping the dpi half in `extract_fields` fails `test_extract_fields_carries_output_type_and_resolution` and `test_extract_fields_defaults_the_resolution_for_pdf`. Hardcoding the resolver to the 200 fallback also fails.

src/Rest_API/Service_Factory.php
tommyshellberg: Fixed in `c0a41c1`. There is now a private `$return_label_v4_memo`, matching `label_service()`, and `$v4_services` is left alone.

Pinned by `test_return_label_service_call_does_not_register_as_injected`, which reflects into `$v4_services` and asserts the `return_label` key is absent after the service has been built. Writing the self-built instance back into that array fails it.

Three more tests came with it: the self-built service is the real `V4\Returns\Service`, it is memoised across calls, and a later `inject_v4_service()` still wins.

tests/php/unit/Rest_API/V4/Returns/Request_BuilderTest.php
tommyshellberg: Fixed in `ea27e14`. All four survivors are dead.

- `print_method`: added `test_known_print_method_is_passed_through` using `consumerPrint`. Hardcoding the resolver to Retail now fails.
- `country`: added `test_country_is_read_from_input_and_uppercased` with a lowercase `be` sender. That one test kills both the hardcoded-NL mutation and the deleted `strtoupper`.
- `output_type`: the fixture is now `JPG`, so deleting `strtolower` fails `test_builds_retail_print_return`.

Also added `test_sender_contact_carries_every_field` — only the email was asserted, so the name and phone could be dropped silently.

src/Rest_API/V4/Returns/Service.php
tommyshellberg: Fixed in `ea27e14`. There is now a `ServiceTest` with twelve tests covering eligibility both ways, the guard order, field extraction, the injected key reaching the wire, and the failure log firing on failure and staying quiet on success.

The two mapper gaps are closed too: `test_first_return_item_returns_the_first_of_several` (every previous fixture held one item, so returning the last was invisible) and `test_decode_content_rejects_invalid_base64` (both fixtures were valid base64, so dropping the strict flag was invisible — and without it a corrupt payload gets written to disk as a broken label instead of being skipped).

I reran the full mutation set afterwards: 20 mutations, 20 killed, no survivors.

src/Rest_API/V4/Returns/Response_Mapper.php
tommyshellberg: Fixed in `c0a41c1` — both methods and their test are deleted.

Nothing called them, and with eligibility now restricted to Dutch customers a partner reference can never be issued on this path. If international returns are built later they can come back with a caller.

The note about the merge step stands and is worth keeping in mind: `maybe_merge_labels()` rebuilds the record with only `type`, `barcode`, `created_at`, `filepath` and `merged_files`, so any key added in `to_label_record()` has to be re-applied afterwards. `api_version` already is.

src/Rest_API/V4/Returns/Service.php
tommyshellberg: Left open on purpose — this one needs PostNL, not a code change.

To restate it with the numbers: `Settings::get_printer_type()` can only produce four strings, which now map to `pdf` (the default setting), `jpg`, `gif` and `zpl`. `png` is unreachable. The SDK comments its own enum case as `// ZPL: not applicable to return endpoints`, and PostNL's guidance says retailPrint output should be PNG or JPG. So the default setting is one PostNL did not name, and one of the four is one the SDK says does not apply.

I have not changed the behaviour, because guessing here could be worse than the current state — mapping PDF to JPG would override a merchant's explicit choice on nothing more than an inference.

Recorded as Q16 in the flip checklist, marked pre-flip. If the endpoint rejects PDF, the default configuration breaks every return label, which is the worst case to find in production. Needs a sandbox key to settle.

Unrelated to this, the resolution half of the same setting was being dropped entirely — fixed in `c0a41c1`, see the Request_Builder thread.

tommyshellberg:
## Fix round done

I've applied the fixes rather than sending them back. Two commits, both green, every fix mutation-verified.

**`c0a41c1` — five defects**

1. **Eligibility read the wrong address.** `is_eligible()` checked the store's country. The print method depends on where the parcel is handed in, which is the customer's. Now reads `$item_info->receiver['country']`. This also makes the EU and ROW merge problem unreachable, since those destinations now fall back to legacy.
2. **`Item_Info` was built before the checkbox was checked**, and it throws `"Barcode is empty!"` when `main_barcode` is absent. Guard moved first, matching the legacy pipeline.
3. **The merchant's print resolution was dropped.** There was already a shared helper for this — `V4\Label\Request_Builder::printer_type_to_label_settings()` splits the setting into format *and* dpi, and the label path has used it all along. The returns service had reimplemented only the format half. It now calls the shared helper and the duplicate loop is gone.
4. **Nothing was logged.** The constructor now requires the client factory, the API key and a PSR-3 logger, matching `V4\Label\Service`. The transport moved into a protected `generate_return()` that logs the original SDK cause. The `method_exists( get_api_key_new )` guard and its empty-key fallback are deleted.
5. **The self-built service was cached in `$v4_services`**, which means "deliberately injected" and is what `barcode_from_label()` reads. It has its own memo now.

`get_partner_barcode()` and `get_partner_id()` are deleted — no caller, and with NL-only eligibility a partner reference can never be issued.

**`ea27e14` — tests**

`Service.php` had none; eleven mutations survived in it. There's now a `ServiceTest` with twelve tests. Three fixtures sat on the value their resolver falls back to, so the resolver could be replaced by a constant with the suite still green — `print_method`, `country` and `output_type` are all pinned now. Plus `first_return_item()` returning the last item, `decode_content()` losing its strict base64 flag, and the sender contact fields beyond the email.

## Left open on purpose

**Which output types the return endpoint accepts.** The default printer setting is PDF, PostNL's guidance says PNG or JPG for retailPrint, and the SDK comments its own ZPL case as not applicable to return endpoints. I have not changed this, because mapping PDF to JPG would override a merchant's explicit choice based on an inference. It needs a sandbox key. Recorded as Q16 in the flip checklist, marked pre-flip.

**The commented-out call site.** `maybe_create_return_label()` is only called from `Order/Base.php:655`, inside a `/* Temporarily commented. */` block, so neither the legacy nor the V4 return-create path runs today. Restoring it is a live behaviour change well outside this PR. Recorded as W25 in the flip checklist, and it means step 4 of the test plan above cannot be run as written.

Everything else from the review is fixed.

tommyshellberg (approved):

PR #350, opened 2026-07-29
23. V4 Letterbox — domestic NL 24h letterbox on the SDK label flow

sal4sup:
### All Submissions:

Task 23 — bring the **domestic NL 24h Letterbox** (mailbox parcel `2928`) onto the V4 label flow.

As the task predicted, letterbox "falls out of the label service": the V4 `Request_Builder` already maps a `letterbox` shipment type to `ShipmentType::LetterBox`, and `V4_Mapper` row 12 already resolves the collapsed `letterbox` option to `2928` / shipment type `letterbox`. The only thing keeping it on legacy was `V4\Label\Eligibility::is_eligible()`, which accepted `parcel` only.

- `Eligibility::is_eligible()` now also accepts the `letterbox` shipment type, **restricted to domestic NL** (parcels may still be domestic or EU/ROW international; letterbox may not).
- No separate barcode call and no meta-shape change: letterbox already rides `Order\Base::create_label()` → `label_service()` and its label is stored under type `label`, exactly as the legacy path does today (the standalone `maybe_create_letterbox()` entry point is not used by the save flow). Because it shares that entry point, it is gated by the existing `postnl_sdk_enable_label` flow flag.
- The **48h Letterbox (`2948`)** stays on legacy automatically: it collapses onto the same `letterbox` option key but keeps product code `2948`, so the mapper's product-code mismatch guard drops it to `has_v4_equivalent = false`.

Both gates (a validated New API Key **and** `postnl_sdk_enable_label`) default off, so merging changes nothing for merchants.

Closes # . — ClickUp: https://app.clickup.com/t/868jqaqhm

Automated (runnable now, no Docker/sandbox needed — unit suite is pure PHP):

1. `vendor/bin/phpunit -c phpunit-unit.xml.dist --filter 'EligibilityTest|Request_BuilderTest'` — new cases: 24h letterbox is eligible, 48h (2948) falls back, an international letterbox is rejected, and the builder maps `letterbox` → `ShipmentType::LetterBox`.
2. `vendor/bin/phpunit -c phpunit-unit.xml.dist` — full suite stays green (499 tests).
3. `composer run check-php` — WPCS clean.

The unit tests drive `resolve_mapped()` with the resolved product code directly; the runtime `_postnl_letterbox_type` meta → `get_product_code()` → `2948` linkage that makes a 48h order fall back is covered by the existing integration test `tests/php/integration/LetterboxTypeTest.php`.

Manual (needs a validated New API Key + PostNL sandbox):

1. On a NL store, enter and validate a New API Key, then set `add_filter( 'postnl_sdk_enable_label', '__return_true' );`.
2. Place/prepare a NL order that qualifies for auto-letterbox (all items letterbox-eligible, combined weight ≤ 2 kg) with the letterbox product default set to **24h**.
3. Generate the label from the order meta box. Expect: a letterbox label PDF stored as before, the barcode captured from the response, and `_postnl_order_metadata['labels']` tagged `api_version = v4` — identical shape to V1.
4. Repeat with the default set to **48h** and confirm it still generates via the legacy path (no `api_version = v4` tag).

### Other information:

### Changelog entry

> Enable the PostNL V4 SDK for the domestic NL 24h Letterbox label (behind the default-off label flag); the 48h Letterbox remains on the legacy API.

### DIFF
--- a/src/Rest_API/V4/Label/Eligibility.php
+++ b/src/Rest_API/V4/Label/Eligibility.php
@@ -17,10 +17,11 @@ if ( ! defined( 'ABSPATH' ) ) {
 }
 
 /**
- * Pure decision logic for whether an order is a parcel the V4 label service
- * handles — a domestic NL parcel or an EU/ROW international parcel from NL/BE.
- * Kept free of WooCommerce and Order\Base so the gate — the highest-risk part
- * of the flow — can be asserted in isolation.
+ * Pure decision logic for whether an order is a shipment the V4 label service
+ * handles — a domestic NL parcel, a domestic NL letterbox (mailbox parcel), or
+ * an EU/ROW international parcel from NL/BE. Kept free of WooCommerce and
+ * Order\Base so the gate — the highest-risk part of the flow — can be asserted
+ * in isolation.
  *
  * @since   6.0.0
  * @package PostNLWooCommerce\Rest_API\V4\Label
@@ -55,8 +56,9 @@ class Eligibility {
 	}
 
 	/**
-	 * Decide whether the collected signals describe a parcel the V4 service handles
-	 * — a domestic NL or EU/ROW international parcel, single- or multi-collo.
+	 * Decide whether the collected signals describe a shipment the V4 service
+	 * handles — a domestic NL or EU/ROW international parcel (single- or
+	 * multi-collo), or a domestic NL letterbox (mailbox parcel 2928).
 	 *
 	 * @param array $signals {
 	 *     Signal set assembled by Service::gather_signals().
@@ -108,11 +110,23 @@ class Eligibility {
 		// equivalent, its services array is the full V4 representation of every
 		// selected option, so no separate product-option gate is needed. Only a
 		// pickup DeliveryLocation is excluded here — that variant lands separately.
-		$mapped = $signals['mapped'] ?? array();
+		$mapped        = $signals['mapped'] ?? array();
+		$shipment_type = (string) ( $mapped['shipmentType'] ?? '' );
 
-		return ! empty( $mapped['has_v4_equivalent'] )
-			&& 'parcel' === ( $mapped['shipmentType'] ?? '' )
-			&& empty( $mapped['deliveryLocation'] );
+		if ( empty( $mapped['has_v4_equivalent'] ) || ! empty( $mapped['deliveryLocation'] ) ) {
+			return false;
+		}
+
+		// Letterbox (mailbox parcel 2928) is a domestic-NL-only variant. The 48h
+		// variant (2948) never reaches here: it collapses onto the same 'letterbox'
+		// option key but keeps product code 2948, so the mapper's product-code
+		// mismatch guard drops it to has_v4_equivalent = false above. A parcel may
+		// be domestic or EU/ROW international.
+		if ( 'letterbox' === $shipment_type ) {
+			return $is_domestic;
+		}
+
+		return 'parcel' === $shipment_type;
 	}
 
 	/**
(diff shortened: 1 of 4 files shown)


tommyshellberg (approved):
Approving. I found no problems in the code. This is the first PR in the V4 queue where every mutation I ran against the tests was caught.

Terms used below: a letterbox parcel is a small shipment that fits through the mail slot (PostNL product code 2928 for 24-hour delivery, 2948 for 48-hour). "Legacy" means the current V1 API path that all merchants use today.

**What this PR does.** It lets the V4 label service accept letterbox shipments, restricted to NL-to-NL. That is the whole production change — about 15 lines in `Eligibility.php` plus doc comments. The request builder and the product matrix already knew about letterbox; only the eligibility gate was blocking it.

**Branch note.** I rebuilt the branch on top of the merged main (it forked from the multi-collo branch's old line). Only this PR's 3 commits were replayed; the diff GitHub shows is now the true scope, 4 files. The description says the full suite is 499 tests — after the rebase it is 661, all green, and CI passes on both suites.

**The description's claims all check out.** I verified each one against the code:
- The standalone letterbox pipeline (`Order\Base::maybe_create_letterbox()` → `Letterbox\Client`) really has no live caller. The "Print Letterbox" button in the order screen is commented out. Real letterbox labels already go through the ordinary shipping label call with product code 2928 or 2948.
- The 48h letterbox stays on legacy through two independent checks: the matrix lists `letterbox_48` as legacy-only, and a 2948 product code fails the product-code comparison even on the shared `letterbox` option key. Both are pinned by tests.
- The SDK enum value is exactly `'letterbox'`, matching the matrix string and the new builder test.

**Mutation testing: 9 of 9 caught.** Every mutation I applied to the changed logic failed at least one test, with different tests catching opposite mutations:
- Hardcoding the letterbox decision to `true` fails the international-letterbox test; hardcoding it to `false` fails the two eligibility tests.
- Each half of the compound guard (V4-equivalent flag, pickup location) is caught by its own data-provider row.
- Disabling the matrix's product-code comparison fails the 48h fall-back test — so that test pins the real mechanism, not just the mapper's own tests.
- Forcing the builder's shipment type to Parcel fails both builder tests.
No test in this PR passes for the wrong reason. The fixture trap that appeared in the previous five PRs (a fixture equal to the code's fallback value) is absent here.

**Parity with the legacy letterbox path holds everywhere I checked.** Every letterbox-specific decision lives in shared code that both paths run: the 24h/48h variant resolution, the forced PDF printer type for letterbox, the 2 kg weight cap (throws before any API call on both paths), the suppressed return barcode, and every settings read. The barcode request has no letterbox branch at all — a letterbox order gets the same 3S barcode as a plain parcel on both paths. Product code 2929 appears nowhere in the codebase, so the portal table's second mailbox code is unreachable from this plugin. Failures are loud: a rejected V4 request is converted, logged with the order number, and shown to the admin in the meta box.

**Two questions for the sandbox, neither blocking (recorded in the flip checklist):**
1. Multi-collo letterbox. The admin screen does not stop you combining a collo count above 1 with letterbox. V1 would send N letterbox shipments; V4 now builds the same request shape. Whether the V4 API accepts it is unknown. V1's own behavior here is equally unverified, so this is parity, but worth one sandbox call.
2. The document type string the V4 response uses for a letterbox label. Both paths store whatever comes back, so a surprising value only changes a label record's inner `type` field, but it should be confirmed once a sandbox key exists.

**One thing for whoever flips flags later.** Letterbox is gated by `postnl_sdk_enable_label` — the same single flag that now covers domestic parcels, service options, EU/ROW international, and multi-collo. There is also a `postnl_sdk_enable_letterbox` flag in `Service_Factory::letterbox_service()`, but it gates only the dead pipeline and does nothing. Turning on the label flag turns on all of it at once. Recorded in the flip checklist.

Remaining before squash: the staging test that proves V1 is unchanged (no real V4 key exists, so that is the only thing staging can prove).

PR #351, opened 2026-07-29
24. V4 Smart Returns (retailPrint) — digital printcode in the customer email

sal4sup:
### All Submissions:

Task 24 — bring **Smart Returns** onto the V4 SDK (`return/generate`, `printMethod = retailPrint`), building on Task 22's unified V4 Returns service.

- **New `V4\Returns\Smart_Returns_Service`** — implements `Smart_Returns_Service_Interface`, reuses Task 22's `Request_Builder`/`Response_Mapper`, and returns the auto-issued barcode plus the retail printcode **image** (forced to PNG). A non-NL order falls back to the untouched legacy Smart Returns pipeline (consumerPrint is a BE follow-on). It does not extend `Order\Base` or touch order meta — a Smart Return is email-only.
- **Factory** — `smart_returns_service()` self-instantiates the V4 service behind the default-off `postnl_sdk_enable_smart_returns` flag (mirrors `label`/`return_label`).
- **Email** — the customer email now shows the printcode/barcode **inline in the body** (embedded via a `cid:` image on `phpmailer_init`) instead of the legacy `PrintcodeLabel` **PDF attachment**, per PostNL: *"do not send a pdf with the label to the consumer, only the printcode."* `codingText` is a routing code (not consumer-facing), so it is not shown.
- **Admin AJAX** (`postnl_send_smart_return_email`) — branches on the response: V4 → printcode in body (no attachment); legacy → PDF attachment path preserved unchanged.
- No `returnPeriod` (V4 default 35 days) and no return services, per Task 22.

Both gates (a validated New API Key **and** `postnl_sdk_enable_smart_returns`) default off, so merging changes nothing for merchants.

Closes # . — ClickUp: https://app.clickup.com/t/868jqaqht

Automated:

1. `vendor/bin/phpunit -c phpunit-unit.xml.dist --filter 'Smart_Returns_ServiceTest|Service_FactoryTest'` — `map_fields()` produces a retailPrint + PNG request with the consumer as sender and the reply-number/return-to-home receiver; the factory self-instantiates the V4 service behind the flag.
2. `vendor/bin/phpunit -c phpunit-integration.xml.dist --filter 'SmartReturnEmailTest'` — the V4 printcode renders inline (`cid:`) with no attachment; the legacy path attaches the PDF and embeds no image; `embed_printcode_image()` adds an inline cid attachment.
3. `vendor/bin/phpunit -c phpunit-unit.xml.dist` — full unit suite stays green (472 tests).

The DTO layer (retailPrint request, reply-number vs home, output types, barcode auto-issue) is already covered by Task 22's `V4\Returns\Request_BuilderTest`/`Response_MapperTest`.

Manual (needs a validated New API Key + PostNL sandbox):

1. On a NL store, enter + validate a New API Key, then `add_filter( 'postnl_sdk_enable_smart_returns', '__return_true' );`.
2. Open a NL order with a PostNL barcode, click **Send email with Smart Return**.
3. Expect: the customer receives the PostNL Smart Returns email with the barcode/printcode shown **inline in the body** (PNG), and **no PDF attachment**. The consumer shows it at a PostNL point.
4. Repeat on a non-NL order (or with the flag off) and confirm the legacy PDF-attachment behaviour is unchanged.

### Other information:

### Changelog entry

> Generate PostNL Smart Returns through the V4 SDK (retailPrint) and show the return printcode inline in the customer email instead of a PDF attachment (behind the default-off Smart Returns flag).


src/Rest_API/V4/Returns/Smart_Returns_Service.php
```diff
@@ -0,0 +1,237 @@
...
+	 * @var Client_Factory|null
+	 */
+	private $client_factory;
+
+	/**
+	 * Service constructor.
+	 *
+	 * @param Client_Factory|null $client_factory Optional SDK client factory.
+	 */
+	public function __construct( ?Client_Factory $client_factory = null ) {
```
tommyshellberg: This constructor is the shape the last four V4 services started with and each one was changed away from it during review (#337, #339, #343, #349). The factory is optional, the key is looked up inside the service through a `method_exists` guard, and nothing here can log — `grep -c logger` over this file is 0, while the error messages `Exception_Converter` produces tell the merchant to check the PostNL logs, which will be empty.

The settled shape, from the sibling `V4\Returns\Service` on main: require `Client_Factory` + the key (marked `#[\SensitiveParameter]`) + a PSR-3 logger in the constructor, delete the `method_exists` guard, and log the original SDK failure in the catch with the order number. I will push this.

src/Rest_API/Service_Factory.php
```diff
@@ -308,7 +309,12 @@ public function postcode_check_service(): Postcode_Check_Service_Interface {
 	 * @return Smart_Returns_Service_Interface
 	 */
 	public function smart_returns_service(): Smart_Returns_Service_Interface {
-		if ( $this->should_use_v4( 'smart_returns' ) && isset( $this->v4_services['smart_returns'] ) ) {
+		if ( $this->should_use_v4( 'smart_returns' ) ) {
+			// A test double injected via inject_v4_service() wins; otherwise build the real V4 service,
+			// which handles the NL retailPrint Smart Return and falls back to the legacy pipeline for the rest.
+			if ( ! isset( $this->v4_services['smart_returns'] ) ) {
+				$this->v4_services['smart_returns'] = new V4_Smart_Returns_Service();
```
tommyshellberg: This writes the self-built service into `$v4_services`. That array's own docblock (twice, lines 96-113) says it means "a V4 service was explicitly injected for this flow" — the label and return-label methods keep their self-built instances in separate `*_v4_memo` properties for exactly that reason. The PR description says this mirrors label/return_label, but it mirrors their pre-fix version; the merged versions on main both use the separate slot.

Nothing crashes today (`barcode_from_label()` only reads the `'label'` key), but the next predicate written against `$v4_services` gets a wrong answer for this flow. Fix: a `$smart_returns_v4_memo` property, same as the other two. I will push this.

src/Rest_API/V4/Returns/Smart_Returns_Service.php
```diff
@@ -0,0 +1,237 @@
...
+	 *
+	 * @param string $output_type One of png|jpg|gif|pdf.
+	 * @return string
+	 */
+	private static function output_type_to_mime( string $output_type ): string {
+		$map = array(
+			'png' => 'image/png',
+			'jpg' => 'image/jpeg',
+			'gif' => 'image/gif',
+			'pdf' => 'application/pdf',
```
tommyshellberg: If PostNL answers with a PDF instead of the requested PNG, this line mails it to the consumer. The response's output type is trusted as-is: `'pdf'` maps to `application/pdf`, the bytes get embedded as an inline email part, the `` tag cannot render a PDF, and most mail clients then show the unreferenced part as a downloadable attachment. That is exactly what PostNL told us not to do — the docblock at the top of this file quotes the instruction. The admin sees a green success and nothing is logged.

Whether the return endpoint can actually answer with a PDF is an open question (Q16 from the task-22 review — nobody has a sandbox key yet), so this path has to be treated as reachable. ZPL is worse: it has no map entry, so raw ZPL text gets embedded as `image/png`.

Fix: accept only png/jpg/gif, and throw (with the actual output type in the message and the log) for anything else. I will push this.

src/Rest_API/V4/Returns/Smart_Returns_Service.php
```diff
@@ -0,0 +1,237 @@
...
+
+		$item = Response_Mapper::first_return_item( $response );
+
+		if ( null === $item ) {
+			throw new \Exception(
+				esc_html__( 'Cannot create the Smart Return. No shipment was returned by PostNL.', 'postnl-for-woocommerce' )
+			);
+		}
+
+		foreach ( Response_Mapper::get_labels( $item ) as $label ) {
```
tommyshellberg: This loop returns the first label that has content, and never reads `labelType`. If the response ever carries two documents — say the return label and the printcode — the consumer may be emailed the wrong one, the admin sees success, and nothing anywhere records that a second label existed. The SDK's `Label` object carries a `labelType` string and the V4 label service already reads it elsewhere.

We do not know the real `labelType` values for this endpoint yet (sandbox question), so selecting by type would be guesswork. What I will push: log a warning with the count and the `labelType` values whenever more than one usable label arrives, keep taking the first, and leave picking by type as a flip-checklist item once sandbox shows the actual strings.

src/Rest_API/V4/Returns/Smart_Returns_Service.php
```diff
@@ -0,0 +1,237 @@
...
+
+			if ( '' === $content ) {
+				continue;
+			}
+
+			$output_type = self::label_output_type( $label );
+
+			return array(
+				'api_version' => 'v4',
+				'barcode'     => Response_Mapper::get_barcode( $item ),
```
tommyshellberg: If the response omits both barcode fields, this quietly becomes an empty string: the email then shows the image but no text barcode underneath it, and a consumer whose mail client blocks images has nothing usable at all. No log at any layer. A Smart Return response with no barcode number is almost certainly a bad response — I will push a logged warning for this case.

src/Emails/WC_Email_Smart_Return.php
```diff
@@ -31,7 +31,39 @@ class WC_Email_Smart_Return extends WC_Email {
...
+
+		/**
+		 * Raw (decoded) printcode image bytes to embed inline in the email body.
+		 *
+		 * Set for the V4 retailPrint Smart Return, whose printcode is shown in the
+		 * body instead of a PDF attachment. Empty for the legacy PDF path.
+		 *
+		 * @var string
+		 */
+		public string $printcode_content = '';
```
tommyshellberg: The four new typed properties add four PHPCS errors under the project ruleset (`composer run check-php` — the PHPCompatibility sniff flags typed properties). CI does not run PHPCS, so this does not fail a check, but the other new V4 files keep the file clean by using docblock-typed untyped properties. I will push that change (types moved to the docblocks, initializers kept).

tests/php/unit/Rest_API/V4/Returns/Smart_Returns_ServiceTest.php
```diff
@@ -0,0 +1,132 @@
...
+
+	/**
+	 * The merchant store info as Smart_Returns\Item_Info::$store exposes it.
+	 *
+	 * @return array
+	 */
+	private function store(): array {
+		return array(
+			'company' => 'My Shop',
+			'country' => 'NL',
```
tommyshellberg, line 184: This fixture makes one of the receiver assertions unable to fail: `'country' => 'NL'` equals the `?? 'NL'` fallback in `map_fields()`, so replacing `$store['country'] ?? 'NL'` with a hardcoded `'NL'` keeps the whole suite green (mutation-verified). This is the sixth PR in this queue where a fixture sits exactly on the fallback value. A `'BE'` store case pins it — I will push one, along with an assertion on `weight_gr` (currently nothing asserts it) and unit tests for `generate()` itself using the fake-HTTP-client helpers task 22 left on main.

src/Rest_API/V4/Returns/Smart_Returns_Service.php
```diff
@@ -0,0 +1,237 @@
...
+ * non-NL origin (consumerPrint is a BE-origin follow-on) falls back to the
+ * untouched legacy Smart Returns pipeline. Because both gates (a validated V4 key
+ * and the per-flow flag) default off, merging this changes nothing for merchants.
+ *
+ * Unlike the outbound return-label service this does not extend Order\Base or
+ * write to order meta: a Smart Return produces an email-only printcode, exactly
+ * as the legacy Smart Returns path did (it wrote a throwaway printcode file just
+ * to attach it).
+ *
+ * @since   5.9.10
```
tommyshellberg: Two doc nits, will fix both: `@since 5.9.10` — the queue settled on `6.0.0` for all V4 SDK work (#343 review). And `Smart_Returns_Service_Interface`'s docblock still documents the raw V2.2 response as the only return shape; it needs a paragraph on the V4 normalized array (`api_version`/`barcode`/`content`/`output_type`/`mime`) and the `api_version` key the caller branches on.

tommyshellberg (commented):
Terms used below: the **printcode** is the barcode image a consumer shows at a PostNL point, where the actual label gets printed. **retailPrint** is the V4 request field that asks for that flow.

The core of the PR is right. The request maps correctly (retailPrint, PNG, no pre-issued barcode, no returnPeriod), the sender/receiver swap matches the legacy V2.2 request field for field, and the email changes are additive — the legacy PDF-attachment path renders exactly as on main. I checked the settings the legacy flow reads: everything that still matters on V4 is inherited through the reused `Item_Info`, and the dropped ones (`location_code`, store email, product codes 2285/3285, the V2.2 printer type) each have a V4 replacement or no V4 field to go to. The legacy Smart Returns flow fires no filters or actions, so nothing needed carrying forward. One nice side effect: on main, a legacy response without a `PrintcodeLabel` entry fatals (undefined `$file_path` assigned to a typed string property — a TypeError the `catch ( \Exception )` misses). This branch replaces that with a clean error message.

**Mutation testing: 8 of 13 killed.** 5 survivors: the receiver-country fixture trap (comment on the test file), the weight (nothing asserts it), and three mutations inside `generate()` — which has zero unit tests, so its response handling, MIME mapping, and the NL routing gate can all be broken without a test noticing. Main already has the fake-HTTP-client helpers that make this testable (task 22's ServiceTest).

**For the flip checklist (not this PR):** which output types `return/generate` actually produces for retailPrint, and whether it can return more than one document per item (and with which `labelType` strings), both need a sandbox key — same family as the existing Q16.

src/Rest_API/V4/Returns/Smart_Returns_Service.php
tommyshellberg: Fixed in f691086: required Client_Factory + SensitiveParameter key + required PSR-3 logger, method_exists guard deleted, transport extracted into a protected generate_smart_return() that logs the original SDK failure with the order number. Mutation-verified both ways: silencing the failure log fails test_failed_call_is_logged_with_the_order_number; logging on success fails test_successful_call_is_not_logged.

src/Rest_API/Service_Factory.php
tommyshellberg: Fixed in f691086: the self-built instance now lives in a smart_returns_v4_memo property, mirroring return_label_service(). An injected test double still wins. The existing factory tests already pin memoisation and the flow name.

src/Rest_API/V4/Returns/Smart_Returns_Service.php
tommyshellberg: Fixed in f691086: the response is accepted only when it carries a png/jpg/gif image (PRINTCODE_MIME_TYPES). A pdf or other non-image response logs the returned output types and throws, so the admin sees a real error instead of a green success plus an unusable email. Pinned by test_pdf_only_response_is_rejected_and_logged; letting the allowlist pass everything or nothing fails different tests.

src/Rest_API/V4/Returns/Smart_Returns_Service.php
tommyshellberg: Fixed in f691086: when more than one usable label arrives (or any label was rejected), a warning is logged naming each labelType and the first usable one is emailed. Picking by labelType stays a flip-checklist item until sandbox shows the real strings. Pinned by test_multiple_usable_labels_warn_and_the_first_wins, and the warning cannot fire on a clean single-label response (test_successful_call_is_not_logged).

src/Rest_API/V4/Returns/Smart_Returns_Service.php
tommyshellberg: Fixed in f691086: a response with no barcode number now logs a warning with the order number; the email still goes out with the printcode image. Pinned both ways by test_missing_barcode_warns and test_successful_call_is_not_logged.

src/Emails/WC_Email_Smart_Return.php
tommyshellberg: Fixed in abfa3ea: types moved to the docblocks, initializers kept. The file is now at 2 PHPCS errors, below main's 3 for the same file (the remaining two are the pre-existing hook-comment pattern).

tests/php/unit/Rest_API/V4/Returns/Smart_Returns_ServiceTest.php
tommyshellberg: Fixed in 8643ab4: a BE store now pins the receiver country (mutating it to a constant fails test_receiver_country_follows_the_store), weight_gr is asserted, and generate() runs under 11 new tests using the task-22 fake-HTTP-client pattern (routing both ways, jpg MIME, pdf rejection, multi-label warning, missing barcode, empty response, failure logging). Mutation certification after the fixes: 13 of 13 killed, zero survivors.

src/Rest_API/V4/Returns/Smart_Returns_Service.php
tommyshellberg: Fixed in f691086 (@since 6.0.0) and abfa3ea (interface docblock now documents both transport shapes and the api_version key the caller branches on).

tommyshellberg:
Review + fix round done at 8643ab4 (three commits on top of the rebased head; per the current process I pushed the fixes rather than sending them back).

**What changed:**
- f691086 — the service now requires the client factory, the API key and a PSR-3 logger (the shape every other V4 service settled on), logs failed calls with the order number, and only accepts a png/jpg/gif printcode: a pdf or other non-image response now logs the returned types and throws instead of being mailed to the consumer as an attachment. More than one usable label, or a missing barcode number, logs a warning. The factory keeps the self-built service in its own memo slot instead of the injected-services array.
- abfa3ea — email properties untyped for PHPCS (types in docblocks), interface docblock documents both return shapes.
- 8643ab4 — 11 new unit tests drive generate() end to end through fake HTTP clients (task 22's pattern); the store-country fixture no longer sits on the fallback value; weight is asserted.

**Numbers:** unit suite 665 → 676 green. Mutation certification after the fixes: 13 of 13 killed, zero survivors — every guard was mutated in both directions (never fires / always fires) and each direction fails a different named test. PHPCS: every touched file is at or below main's error count; the new V4 files are clean.

**For the flip checklist, needs a sandbox key:** which output types return/generate actually produces for retailPrint, and whether it can return more than one document per item (and with which labelType strings). Until then the code takes the first usable image and logs what else arrived.

One stale line in the PR description: the unit count is now 676, and the factory wiring now uses a separate memo slot (the description's 'mirrors label/return_label' pointed at their pre-review shape).

PR #352, opened 2026-07-30
25. Filter/action carry-forward audit for the V4 paths

sal4sup:
### All Submissions:

* [x] Have you successfully run tests with your changes locally? (Unit suite green — 475 tests; the new integration test runs in wp-env/CI.)
* [ ] Have you tested both blocks/non-blocks cart/checkout? — N/A: admin label-generation path + an internal filter, no checkout surface.
* [ ] Have you tested the cart and checkout as both a logged-in user and a guest? — N/A.
* [ ] Have you successfully placed an order as both a logged-in user and a guest? — N/A.
* [ ] Did you have the query monitor plugin active during all testing? — N/A.

Task 25 — the cross-cutting filter/action carry-forward audit. Ensures the public extension points third parties rely on keep firing from the V4 SDK paths with the **same parameter shape** as legacy.

Audit result — only one gap:

| Extension point | Carry-forward status |
|---|---|
| `postnl_shipment_addresses` | **Was not firing** on V4 — the V4 `Request_Builder` assembles addresses from a flat field array, bypassing `Legacy\Shipping\Client` where the filter fires. **Now fired** from the V4 label path. |
| `postnl_order_weight` | Already carries forward — the V4 label & returns services reuse the shared `Shipping\Item_Info`, which fires it during construction. |
| `postnl_order_meta_box_fields` | Already fires — shared `Order\Base` meta box. |
| `postnl_logger_write_message` | Already fires — shared `Logger`. |

The fix reuses `Legacy\Shipping\Client::get_shipment_addresses()` verbatim so the `( array $addresses, Shipping\Client $client )` shape is byte-identical to legacy, then `Request_Builder::apply_filtered_receiver()` maps the (possibly modified) recipient back onto the request so third-party edits actually reach it. Returns and smart-returns legacy paths never fired `postnl_shipment_addresses`, so their V4 counterparts intentionally don't either (carry-forward = match legacy, not add surface).

Behaviour is unchanged for merchants: both V4 gates (validated New API Key + `postnl_sdk_enable_label`) default off.

Full matrix: `docs/postnl-v4-migration/approach-2/filter-carry-forward-matrix.md`.

ClickUp: https://app.clickup.com/t/868jqaqhv

Stacked on `postnl-v4-smart-returns` (task 24).

1. `npm run test:php:unit` — the new `apply_filtered_receiver()` cases in `Request_BuilderTest` pass (pure recipient-overlay mapping).
2. `npm run test:php:integration` (wp-env) — `FilterCarryForwardTest` asserts `postnl_order_weight` and `postnl_shipment_addresses` fire from the V4 label service's request-construction seam with the documented args, and that a third-party address rewrite is honoured, not fired-and-ignored.
3. Optional manual: with a validated New API Key + `postnl_sdk_enable_label` on, hook `postnl_shipment_addresses` in a snippet, generate a domestic NL label, and confirm the hook fires and an address edit reaches the PostNL request.

### Other information:

### Changelog entry

> Internal (V4 migration, flag-gated, no merchant-facing change): fire `postnl_shipment_addresses` from the V4 label path so third-party address filters carry forward; document the filter/action carry-forward matrix.


tests/php/integration/FilterCarryForwardTest.php
```diff
@@ -0,0 +1,275 @@
...
+	private function make_service(): V4_Label_Service {
+		return new class() extends V4_Label_Service {
+			/**
+			 * Skip the V4 Service constructor.
+			 *
+			 * It requires the SDK client factory, resolved key and logger, but this
+			 * seam exercises only filter_shipment_addresses(), which builds its own
+			 * Legacy\Shipping\Client from the Item_Info and needs none of them.
+			 */
+			public function __construct() {
```
tommyshellberg: This subclass never reaches `create()`. Delete `Service.php:160-163` and all three tests here still pass. I ran it: 3/3 green with the production call removed.

The fix is to drive the real `create()`. `ServiceTest` already has `Spy_Label_Client_Factory` with an injectable fake HTTP client. Build the service with that, a fake key and a `NullLogger`, register the filter, call `create( $post_data )`, and assert on `$http->last_request`: the body must contain `"Rewritten Street"`. That one test proves the filter fires from the V4 path *and* that the edit reaches the wire. It replaces the hand-composed check at line 176.

Do the same for `postnl_order_weight` (assert it fired during `create()`) and add one each for `postnl_order_meta_box_fields` and `postnl_logger_write_message` from a V4 service. The task says every public filter. The logger one will fail today; see the matrix comment.

tests/php/integration/FilterCarryForwardTest.php
```diff
@@ -0,0 +1,275 @@
...
+			$addresses = $this->make_service()->expose_filter_addresses( $this->make_post_data() );
+		} finally {
+			remove_filter( 'postnl_shipment_addresses', $rewrite, 10 );
+		}
+
+		// Compose the two calls create() makes — the filtered addresses through
+		// apply_filtered_receiver() — so the assertion proves the rewrite lands on
+		// the receiver fields the DTO consumes, not merely that add_filter mutated
+		// the array.
+		$receiver = Request_Builder::apply_filtered_receiver(
```
tommyshellberg: This composes the two calls itself, so it proves `apply_filtered_receiver()` works, not that `create()` calls it. Once the test above drives the real `create()`, this test can go.

tests/php/integration/FilterCarryForwardTest.php
```diff
@@ -0,0 +1,275 @@
...
+		try {
+			$this->make_service()->expose_filter_addresses( $this->make_post_data() );
+		} finally {
+			remove_filter( 'postnl_order_weight', $spy, 10 );
+		}
+
+		$this->assertNotEmpty( $captured, 'postnl_order_weight must fire while the V4 service parses the order.' );
+
+		$last = end( $captured );
+		$this->assertIsNumeric( $last[0], 'The first argument must be the numeric total weight.' );
```
tommyshellberg: `assertIsNumeric` accepts `"1.5"` as a string. The documented shape is `float`. Assert `assertIsFloat` and the value (1.5, the product weight), or a weight bug that still returns a number passes.

docs/postnl-v4-migration/approach-2/filter-carry-forward-matrix.md
```diff
@@ -0,0 +1,69 @@
...
+
+Two gates still guard every V4 path (a validated New API Key *and* the per-flow
+`postnl_sdk_enable_{flow}` flag, both default off), so this changes nothing for
+merchants until a flow is turned on.
+
+## Matrix
+
+| Extension point | Legacy fire point | Parameter shape | V4 label | V4 returns | V4 smart returns |
+|---|---|---|---|---|---|
+| `postnl_shipment_addresses` | `Legacy\Shipping\Client::get_shipment_addresses()` | `( array $addresses, Shipping\Client $client )` | **Fired (new)** — `V4\Label\Service::filter_shipment_addresses()` reuses the legacy address builder verbatim, then overlays the filtered recipient back onto the request | n/a — legacy returns never fired it | n/a — legacy smart returns never fired it |
```
tommyshellberg: The returns column is wrong. Legacy return labels do fire this filter. `Legacy\Return_Label\Client` extends `Shipping\Client` and only overrides `get_customer_address()`, so it inherits `get_shipments()` and the `apply_filters()` at `Shipping/Client.php:292`. `Order\Base:1459` builds it for every return label today.

`V4\Returns\Service::create()` never fires it. So when `postnl_sdk_enable_returns` flips, a merchant's address-rewrite plugin stops applying to return labels with no error. Either wire it into the V4 returns service the same way as the label service, or record it in the flip-checklist as a blocker for the returns flag. Either way the row needs correcting.

docs/postnl-v4-migration/approach-2/filter-carry-forward-matrix.md
```diff
@@ -0,0 +1,69 @@
...
+| `postnl_order_weight` | `Legacy\Shipping\Item_Info::calculate_order_weight()` | `( float $total_weight, WC_Order $order )` | Fires — V4 service builds the shared `Shipping\Item_Info` | Fires — `Return_Label\Item_Info` extends `Shipping\Item_Info` | n/a — `Smart_Returns\Item_Info` extends `Base_Info`; legacy never fired it |
+| `postnl_order_meta_box_fields` | `Order\Base::meta_box_fields()` | `( array $fields, string $context )` | Fires — shared admin meta box (not a per-service request step) | Fires — shared | n/a — smart returns is email-only |
+| `postnl_logger_write_message` | `Logger::write()` | `( string $message )` | Fires — shared logger | Fires — shared | Fires — shared |
```
tommyshellberg: This filter does not fire from any V4 service. V4 services log through `SDK\Logger_Adapter::log()`, which calls `wc_get_logger()->log()` directly. The filter is only in `Logger::write()` (`src/Logger.php:57`), which the adapter never calls.

A merchant plugin that redacts or rewrites PostNL log lines sees none of the V4 output. The adapter also skips `check_pdf_content()`, so a V4 log line containing a label PDF goes into the log as raw base64.

Simplest fix: have `Logger_Adapter::log()` call `$this->logger->write( $message )` instead of `wc_get_logger()` directly. That restores the filter and the PDF stripping in one line. Then the row can say "Fires" and a test can prove it.

docs/postnl-v4-migration/approach-2/filter-carry-forward-matrix.md
```diff
@@ -0,0 +1,69 @@
...
+  subclasses it), and weight is computed — and the filter fired — during that
+  construction, before any V4-specific translation. Smart returns uses a
+  `Base_Info` subclass that never computed order weight, so legacy never fired
+  the filter there and neither does V4.
+
+- **`postnl_order_meta_box_fields` and `postnl_logger_write_message` are shared.**
+  They fire from `Order\Base` and `Logger`, which every service (legacy and V4)
+  reuses unchanged, so they carry forward with no V4-specific work.
+
+## Known differences from legacy (in-scope)
```
tommyshellberg: Two more differences belong in this section.

House number. A filter that sets `'Street' => 'Foo 12', 'HouseNr' => ''` (the usual Belgium/Germany pattern) sends `HouseNr: ""` on legacy. On V4, `maybe_null()` at `Request_Builder.php:331` turns `''` into `null` and the SDK omits the field. The SDK `Address` DTO has an `addressLine` field for exactly the combined form; the builder never uses it.

Contact block. Legacy `Contacts` carry only email and phone (`Shipping/Client.php:91-97`). On V4 `Request_Builder::contact()` reads `first_name`, `last_name` and `company` from the same filtered receiver. So a filter that changes the recipient name now changes the contact too, only on V4.

src/Rest_API/V4/Label/Service.php
```diff
@@ -155,15 +156,37 @@ public function create( array $post_data ): array {
...
+	 * builder is reused verbatim rather than reconstructed: the array shape and the
+	 * client argument are exactly the ones the legacy path passes. The returned
+	 * (possibly modified) addresses are overlaid back onto the receiver by the
+	 * caller so third-party edits actually reach the request.
+	 *
+	 * @param Shipping\Item_Info $item_info Parsed legacy item info.
+	 * @return array Filtered legacy-shaped address array.
+	 */
+	protected function filter_shipment_addresses( Shipping\Item_Info $item_info ): array {
+		return ( new Legacy_Shipping_Client( $item_info ) )->get_shipment_addresses();
```
tommyshellberg: A filter that forgets to `return` makes this throw a `TypeError` because of the `: array` return type. `TypeError` is an `\Error`, not an `\Exception`. The AJAX handlers catch only `\Exception` (`Order/Single.php:490`, `Order/Bulk.php:725`), and `admin-order-single.js` has no `.fail()` handler. The merchant sees the label form greyed out with no message. Bulk stops at that order.

Legacy on the same filter sends `Addresses: null`, PostNL rejects it, the merchant sees PostNL's message and the request is in the log.

Check the value and throw a plain `\Exception` naming the filter, so the existing catch blocks show it:

```php
$addresses = ( new Legacy_Shipping_Client( $item_info ) )->get_shipment_addresses();
if ( ! is_array( $addresses ) ) {
	$this->logger->error( 'postnl_shipment_addresses returned ' . gettype( $addresses ) . ' instead of an array.' );
	throw new \Exception( esc_html__( 'A plugin hooked into postnl_shipment_addresses returned an invalid address list.', 'postnl-for-woocommerce' ) );
}
return $addresses;
```

src/Rest_API/V4/Label/Service.php
```diff
@@ -155,15 +156,37 @@ public function create( array $post_data ): array {
 			return $this->create_label_pipeline( $post_data );
 		}
 
-		$fields   = $this->extract_fields( $item_info, $signals['mapped'], $post_data );
-		$request  = Request_Builder::build( $fields );
-		$response = $this->confirm_label( $request, $fields );
+		$fields             = $this->extract_fields( $item_info, $signals['mapped'], $post_data );
+		$fields['receiver'] = Request_Builder::apply_filtered_receiver(
```
tommyshellberg: Eligibility ran at line 155 on the unfiltered address. Nothing re-checks after the overlay. A filter that sets `Countrycode => 'DE'` sends product 3085 (domestic NL parcel) with a German address. A code the SDK enum does not know (`'UK'`, `'NLD'`) is silently reset to NL by `Request_Builder::country()` line 363. Legacy passes the string through and PostNL rejects it.

Cheapest safe option: if `$fields['receiver']['country']` changed from what eligibility saw, log it and fall back to `create_label_pipeline()`. That keeps the legacy behaviour for exactly the orders a filter touches in a way V4 does not handle yet.

src/Rest_API/V4/Label/Request_Builder.php
```diff
@@ -423,6 +423,51 @@ public static function printer_type_to_label_settings( string $printer_type ): a
...
+			'postcode'         => 'Zipcode',
+			'city'             => 'City',
+			'country'          => 'Countrycode',
+		);
+
+		foreach ( $addresses as $address ) {
+			if ( '01' !== ( $address['AddressType'] ?? '' ) ) {
+				continue;
+			}
+
```
tommyshellberg: `(string)` on an array gives the literal string `Array` plus a PHP warning nobody sees. The label prints with a street called "Array". An object without `__toString` throws an `\Error` and hits the same uncaught path as the null return. Legacy sends the nested value and PostNL rejects it with a readable message.

Accept scalars and null only. Throw a plain `\Exception` naming the key for anything else.

tests/php/unit/Rest_API/V4/Label/Request_BuilderTest.php
```diff
@@ -560,4 +560,82 @@ public static function printer_type_provider(): array {
...
 		);
 	}
+
+	/**
+	 * A recipient (AddressType 01) entry in the legacy address shape.
+	 *
+	 * @param array $overrides Fields to override on the base recipient entry.
+	 * @return array
+	 */
+	private function recipient_address( array $overrides = array() ): array {
```
tommyshellberg: This fixture equals `domestic_fields()['receiver']` on all nine mapped fields: `''`, Jan, Jansen, Main Street, 9, A, 1234AB, Amsterdam, NL. Overlaying it is the same as doing nothing, so the preserves-contact test at line 615 passes even if the function overwrote everything.

Mutation results, 6 of 13 caught. These survive: swap `FirstName`/`Name`, drop `company`, drop `house_number_ext`, drop `country`, remove the `break`, `array_key_exists` to `isset`, remove the `(string)` cast.

Give every field here a value that differs from the receiver fixture (`Piet`, `Pietersen`, `Rewritten BV`, `bis`, `BE`, ...) and assert all nine in the overlay test. Add one test with two `01` entries (first must win) and one with `'Street' => null` (must become `''`). That kills the survivors.

tommyshellberg (changes requested):
Requesting changes. The address filter wiring itself is right: reusing the legacy `get_shipment_addresses()` keeps the `( array $addresses, Shipping\Client $client )` shape byte-identical, and I checked that building a `Legacy\Shipping\Client` inside the V4 path is safe (no network call, no validation, no logging; `Rest_API\Base::__construct()` only assigns properties). Hard gates pass: SDK still in `require-dev` at `dev-main#5a064dd`, `php >= 7.4`, `Service_Factory` and `Order\Base` untouched, filter signature unchanged. Both flags default off, so V1 is unaffected.

Three things block merge.

**1. No test proves `create()` fires the filter.** I deleted the four-line call at `Service.php:160-163` and ran the integration suite in wp-env. It stayed green, 3/3. The test subclass skips the constructor and calls `filter_shipment_addresses()` directly. The third test then calls `apply_filtered_receiver()` by hand. So the suite proves each helper works when the test calls it. It does not prove the V4 label path calls them. That is the acceptance criterion for this task, and it is what Thomas asked for: verifying the hooks fire without manual testing. Details inline.

**2. The matrix is wrong on two rows.**
- Legacy return labels *do* fire `postnl_shipment_addresses`. `Legacy\Return_Label\Client` extends `Shipping\Client` and overrides only `get_customer_address()`, so it inherits `get_shipments()` and the filter call at `Shipping/Client.php:292`. `Order\Base:1459` builds that client for every return label. `V4\Returns\Service` never fires it. Once `postnl_sdk_enable_returns` is turned on, a merchant's address-rewrite plugin silently stops applying to return labels.
- `postnl_logger_write_message` does *not* fire from any V4 service. Every V4 service logs through `SDK\Logger_Adapter::log()`, which calls `wc_get_logger()->log()` directly. The filter lives only in `Logger::write()` (`src/Logger.php:57`), which the adapter never calls. The same bypass skips `check_pdf_content()`, which is what keeps base64 PDF blobs out of the log file.

**3. A bad filter return crashes V4 with no message.** `filter_shipment_addresses()` declares `: array`. A filter that forgets to `return` (very common) makes it throw a `TypeError`. The AJAX handlers catch only `\Exception` (`Order/Single.php:490`, `Order/Bulk.php:725`), and `admin-order-single.js` has no `.fail()` handler. The merchant sees the label form greyed out forever. On legacy the same filter sends `Addresses: null`, PostNL rejects it, and the merchant sees PostNL's error in the meta box and the request in the log.

**Mutation results on the new unit tests: 6 of 13 caught.** `recipient_address()` in `Request_BuilderTest` equals `domestic_fields()['receiver']` on all nine mapped fields, so overlaying it is the same as doing nothing. Survivors: swap FirstName/Name, drop `company`, drop `house_number_ext`, drop `country`, remove `break`, `array_key_exists` to `isset`, remove the `(string)` cast. Caught: the no-op return, the `'01'` check, and dropping street/city/house_number/postcode. Fix inline.

**Checked and fine:** `postnl_order_weight` genuinely carries forward on the label and returns paths (both build a `Shipping\Item_Info` subclass that runs `calculate_order_weight()`). `postnl_order_meta_box_fields` fires from shared `Order\Base`. Smart Returns and Shipment_and_Return legacy clients extend `Rest_API\Base`, not `Shipping\Client`, so the "never fired" claim is correct for those two. No other `postnl_*` hook fires on a legacy label or return path. The round trip `extract_fields()` -> legacy address array -> `apply_filtered_receiver()` is a no-op field for field when no filter is registered.

The branch is already rebased cleanly onto main (4 own commits, 5 files). No rebase work needed.

